PR #1, opened 2020-11-11
Feature/helmscript

lukma99:
Checks if (the desired version of) helm is installed, and if not proceeds to do so, if user allows it.


schnatterer:
Thanks!
Helm Looks good.
What about `kubectl`? It's also a prerequisite, isn't it? 
And you told me earlier about an error when no the `~/.kube` diretory doesn exist.

lukma99:
Didn't have time for all the steps yet. If you want you can wait till I finished all of them

lukma99:
@schnatterer 
Added automatic kubectl installation now.

the creation of ~/.kube diretory doesn't seem to be necessary. In my latest tests the script created a kubeconfig, if not present.

PR #9, opened 2020-11-30
feature/implement_resource_validation

phaenr:
Implement k8s and yaml resource validation with kubeval and yamllint in petclinic plain pipeline


phaenr:
closed it because I thought I would have to open a new one, due to a late merge from main back into feature branch. But apparently the PR picks it up to. :)

PR #28, opened 2021-03-22
Feature/update terraform

schnatterer:
Minor changes to terraform.

* Currently fails because `Error: googleapi: Error 400: Master version "1.17.13-gke.2600" is unsupported., badRequest`
* credentials `account.json` is a fixed path, which is CI-unfriendly.


schnatterer:
Closely coupled with `external_scmm` story. Merging into it. See #29.

PR #51, opened 2021-06-24
Jenkins: Fix failing K8s agent builds.

schnatterer:
By upgrading to (almost) latest helm chart version and explicitly installing plugin dependencies.

Note: Didn't upgrade to latest helm chart version because it lead to errors on build, such as the following:
` Remote call to JNLP4-connect connection to jenkins-agent.default.svc.cluster.local/10.43.179.183:50000`

### DIFF
--- a/scripts/jenkins/init-jenkins.sh
+++ b/scripts/jenkins/init-jenkins.sh
@@ -7,7 +7,9 @@ if [[ -z ${PLAYGROUND_DIR+x} ]]; then
   PLAYGROUND_DIR="$(cd "${BASEDIR}" && cd .. && cd .. && pwd)"
 fi
 
-JENKINS_HELM_CHART_VERSION=3.1.9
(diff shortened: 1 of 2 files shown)


pmarkiewka (changes requested):
typo in "...newer agent.tag then we and update"

schnatterer:
Typo fixed

scripts/jenkins/init-jenkins.sh
```diff
@@ -9,6 +9,8 @@ fi
 
 # When Upgrading helm chart, also upgrade additionalPlugins in jenkins/values.yaml
 # Check for "Failed to load" in the Jenkins log and add or upgrade the plugins mentioned there appropriately.
+# In addition, check if the new helm chart version uses a newer agent.tag when we and update.
```
pmarkiewka, line 12: "...when we and update."

schnatterer: I removed the `when we and update` completely, as the whole paragraph is about upgrading.

Issue #55, opened 2021-08-03
argocd/nginx and fluxv1/nginx won't build on remote cluster


marekzan:
error: `hudson.remoting.ProxyException: java.util.concurrent.ExecutionException: Invalid object ID 8 iota=14`

This bug seems to only appear when running the gop through `kubectl run` and on a remote cluster.

schnatterer:
I can confirm that this also occurs when applying the playground via `scripts/apply.sh --remote` locally.

schnatterer:
It might be related to the `timeout` step, which is only used with the `nginx` pipelines, not in the other applications.

schnatterer:
Updating all plugins resolves this issue.
Maybe it's enough to install the latest version of the [basic steps plugin](https://plugins.jenkins.io/workflow-basic-steps/) that contains the timeout?

schnatterer:
It's really odd that so many plugins are installed and most of them are really old versions.

Once more reason why we really should consider building our own Jenkins image as stated in the [helm chart's docs](https://github.com/jenkinsci/helm-charts/tree/17a4ca9fc514817a167e6baa3feb4693d64883c1/charts/jenkins#consider-using-a-custom-image).

schnatterer:
For some reason we end up with way older plugins when using `--remote` then when using a local k3d cluster :thinking: 

Remote:
![image](https://user-images.githubusercontent.com/1824962/128028328-bc3e7ae8-7e1e-4a7e-af6f-a599682cf328.png)
vs.
local:
![image](https://user-images.githubusercontent.com/1824962/128028114-60d8f229-a07e-4f4b-9dfc-5f09cfbe61f6.png)

schnatterer:
I think one reason for this is that our Jenkins REST client does not validate the proper version is set.
It only checks, if the plugin is installed at all:
https://github.com/cloudogu/gitops-playground/blob/47b6272f83fc021a811f3b1bec9fe902d66ef706/scripts/jenkins/jenkins-REST-client.sh#L110-L113

So, I think the most sustainable fix would be to improve this behaviour to verify the proper versions are installed. This should make the whole thing *way* more deterministic.

phaenr:
Occurs also on a local instance running the tldr command.

phaenr:
fixed

PR #60, opened 2021-08-09
removed timeout step so builds won't fail

marekzan:
Signed-off-by: Marek Markiewka <marek.markiewka@cloudogu.com>


schnatterer:
Works around, i.e. fixes #55  (for now).

PR #98, opened 2022-11-10
Vault and External Secrets Operator integration

schnatterer:
* Implements optional Vault and External Secrets Operator integration via a `--vault=dev|prod` parameter.
* Initializes Vault with static secrets in dev mode via post-startup hook
* Adds example stage-specific `SecretStores` to ArgoCD and `ExternalSecrets` to an example app (ArgoCD-only)
* Mounts secret into example app, exposes via HTTP for demo purposes
* Extends docs .Also creates a simpler diagram with all features at the top.   
  Note that the diagrams in README won't show until merged to master (because we use the plantuml proxy for rendering)

Can be tested with preview image:  `ghcr.io/cloudogu/gitops-playground:159954b`

In order to implment this feature properly, some major refactorings needed to be done:
* 2eaee634
  * Simplify package structure of Grovvy code. There were more packages than classes. Well, almost. 2eaee634
  * `modules` were renamed to `features` to better reflect what they are to the GitOps playgrounds, e.g. monitoring and secrets managements are optional features 
   * Disentangled MetricsModule. Mailhog is always on. ArgoCD Notifications is always on. Only Prometheus Stack is enabled/disabled.
* ef07e1ca Mailhog and Prometheus Stack can now be deployed for all GitOps operators 
* 49fe9bd5 Create a declarative default configuration in `ApplicationConfigurator` -> this holds a lot of constants that could be overwritten easily using a config file, no need to add dozens of CLI params. 
* b86fc65b CommandExecutor now fails on error, instead of silently continuing. A bit like `set -o errexit`. This lead to a couple of improvments for idempotence. e.g. where something failed in the past but was never recognized. Example: git push of control app. 
* b4559cd3 Make GitClient more generic: Remove control-app repo specifics, make explicitly stateful. Rename to `ScmmRepo` to better match what it is. A representation of a repostitory in SCMM, not a generic client 
* bfa2f877 Remove code for setting the SCMM default branch 

Minor unrelated fixes
* Jenkins upgrade 4d95f75e
* Avoid CVE-2022-37865 in ivy  e5fe04eb
* Fix app helm-dependency for local deployment 38518132
* Fix example app: nginx-helm-jenkins 0cd0567e
* Add missing README to control-app d9cd9a66
* Add overview of dependencies 103ca342


schnatterer:
* Merged Branch `feature/argocd-2.5` into  `vault`, Otherwise Jenkins and SCMM would no longer have worked. Cherry picking of  5 commits would have been to much for my taste. Also the argocd-2.5 branch has been battle-tested in a training.
* Had to update NGINX Helm Charts, because Bitnami removes all versions older than 6 months from their helm chart index. To make sure we're not forced to update again, switch to full index.
* Switched from using the `root` token to using k8s service accounts and a user account.
  *  For one this allows arbitrary `--password`s.  As the `root` token is also used as an ID, having special character in `--password` resulted in `Error initializing Dev mode: failed to create root token with ID "<PW with special chars>"`.
  * Not using the root token is much more realistic!
  * If necessary, the root token can be found on the log

schnatterer (commented):
Did it just work without `namespaceOverride`?

PR #105, opened 2023-03-30
Push branch "test" as image tag "test" + Get rid of CVE-2022-1471

schnatterer:
The moving tag `test`  allows us to have a constant image tag that can be used to review features in customer environments before merging them to main.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -169,7 +169,8 @@ COPY src /app/src
 USER 0
 # Avoid criticla CVE in ivy, which is not fixed in groovy 3 right now. We don't use trivy anyway, so delete it.
 RUN rm /opt/groovy/lib/ivy-2.5.0.jar
-
+# Avoid criticla CVE in SnakeYAML, which is not fixed in groovy 3 right now. Our app brings its own dependency for snakeyaml
+RUN rm /opt/groovy/lib/snakeyaml-1.30.jar
 
 
 # Pick final image according to build-arg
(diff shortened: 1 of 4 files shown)


schnatterer:
Unrelated to this PR the sudden occurrence of CVE-2022-1471 made the build unstable in `dev` image.

Best solution at the moment:
*  Groovy dependency removed from dev image
* App depedencies fixed by upgrading micronaut and explicitly adding `jackson-dataformat-yaml` (only available as RC 😬) which is a dependency of `groovy-yaml` which in turn had the vulnerable snakeyaml dependency 🥵

We just have to remember to replace the RC with the proper version once it is released, and take out the dependency later when upgrading Groovy for the next time.

schnatterer:
After deleting snakeyaml from dev image I came to the realization, that  `gitops-playground.jar` already contains groovy-all, so we shouldn't need a groovy base image.
Switched to JDK reduced local (uncompressed) image size 700 -> 640 MB.
Then I realized that even a JRE should do. Further reduced local (uncompressed) image size 640 -> 450MB.
This also reduces the number of dependency, i.e. attack surface, i.e. probability that the CVE's occur in the dev image.
Another nice perk is that we no longer need to maintain a redundant groovy version in Dockerfile.

pmarkiewka (approved):
lgtm

schnatterer:
BTW - compressed images sizes before and after this PR:

```
❯ docker-image-size  ghcr.io/cloudogu/gitops-playground:3526dd6-dev

ghcr.io/cloudogu/gitops-playground:3526dd6-dev: 462 MB
❯ docker-image-size  ghcr.io/cloudogu/gitops-playground:f2fece2-dev

ghcr.io/cloudogu/gitops-playground:f2fece2-dev: 271 MB
```

PR #107, opened 2023-04-06
add use of prebuild jenkins image

movray:


schnatterer:
https://github.com/cloudogu/jenkins-helm-image

schnatterer:
This PR is supposed to resolve failing starts of the playground caused by changing Jenkins plugins (transitive) dependencies.
So from now on each version of the GitOps playground should run almost deterministicly. At leas in regard to Jenkins. For SCM-Manager we still have a an issue with latest plugins being installed. But it rarely causes trouble.

Issue #108, opened 2023-04-28
Spring Petclinic repo builds failing


schnatterer:
` unable to download https://repo.spring.io/release/org/springframework/cloud/spring-cloud-bindings/1.8.0/spring-cloud-bindings-1.8.0.jar` 

Probably caused by this: https://spring.io/blog/2022/12/14/notice-of-permissions-changes-to-repo-spring-io-january-2023

See also:       https://github.com/spring-projects/spring-boot/issues/33699

Possible workaround? https://github.com/spring-projects/spring-boot/issues/33699 Does not seem to work.

Fix: Update Base to latest [0.3.229-base](https://hub.docker.com/layers/paketobuildpacks/builder/0.3.229-base/images/sha256-f59dca3a71563278db098a2d249e22866deb4e52ad756f223b74455e1bc2a3f6?context=explore) in `Jenkinsfile`s

PR #111, opened 2023-05-12
Added creator variable to set labels

pwrdeli:
only used if the variable is not null, so we can set it from our Jenkinsjob


schnatterer:
Due to #108 this will always lead to failing builds. 
Cherry picked the fixing commit from #110.

schnatterer:
Cherry picked some more necessary fixes regarding Jenkins+plugins (as usual) and our new Jenkins instance that builds this repo 😬

PR #112, opened 2023-06-14
Removed stale documentation

CloudoguSiebels:


scripts/apply.sh
```diff
@@ -626,8 +626,8 @@ function printWelcomeScreenArgocd() {
   if [[ $INSTALL_ARGOCD == true ]]; then
     echo "| For ArgoCD:"
     echo "|"
-    echo -e "| - GitOps repo: \e[32m${SCMM_URL}/repo/argocd/gitops/code/sources/main/\e[0m"
-    echo -e "| - Pull requests: \e[32m${SCMM_URL}/repo/argocd/gitops/pull-requests\e[0m"
+    echo -e "| - GitOps repo: \e[32m${SCMM_URL}/repo/argocd/cluster-resources/code/sources/main/\e[0m"
+    echo -e "| - Pull requests: \e[32m${SCMM_URL}/repo/argocd/cluster-resources/pull-requests\e[0m"
```
schnatterer, line 630: We talked about this before, but now that I see the diff, I'd like to change `GitOps repo` to `example apps repo` and also change the URL to `example-apps`. This is the repo which is most likely the most interesting for first time users.

We could discuss to also add the `cluster-resources`, and the `argocd` repo. But most likely we will be removing the welcome screen sooner or later. So lets not waste too much time here.

CloudoguSiebels: agree!

PR #116, opened 2023-06-23
Fix broken deployments and exercises

CloudoguSiebels:


CloudoguSiebels:
> First, Jenkins is not picking up the repo automatically, because it's not authorized.

We configure Jenkins to pick up all projects in that namespace. I think it's coherent to grant access to the entire namespace.

Myself, I moved the repository from one namespace to the other and did not run into issues. Except when redeploying the playground iteratively. Then, it is missing the exercise in the exercise repository.
I think mentioning the import feature is a good idea.

> **petclinic-helm OOM**

I think we should fix this in the chart directly. There we explicitly set a value that probably never works anymore.

Additionally pinning the builder version is a good idea as well.

PR #118, opened 2023-06-30
Create a Jenkins API Client in Groovy to create users

CloudoguSiebels:


scripts/utils.sh
```diff
@@ -50,3 +50,27 @@ function spinner() {
 function error() {
      echo "$@" 1>&2; 
 }
+
+# Entry point for the new generation of our apply script, written in groovy
+function runGroovy() {
+
```
schnatterer, line 56: Now that we have several entry points, I'd like to have a log statement with params. However, this would also print secrets. Any ideas?

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
```diff
@@ -0,0 +1,51 @@
+package com.cloudogu.gitops.cli
+
+import com.cloudogu.gitops.dependencyinjection.JenkinsFactory
+import com.cloudogu.gitops.jenkins.UserManager
+import groovy.util.logging.Slf4j
+import io.micronaut.context.ApplicationContext
+import picocli.CommandLine
+
+@CommandLine.Command(name = 'jenkins', description = 'CLI-tool to configure jenkins.',
```
schnatterer, line 9: Any reason why you didn't use imports like  `import picocli.CommandLine.Command ` instead of using qualified references?
IMO the code is easier to read without the repetitive `CommandLine.`

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @CommandLine.Command(name = "set-permission", description = "adds user to jenkins", mixinStandardHelpOptions = true)
+    void grantPermission(
+            @CommandLine.Parameters(paramLabel = "username", description = "The username to grant the permission to") String username,
+            @CommandLine.Parameters(paramLabel = "permission", description = "The permission to grant") String permission,
+            @CommandLine.Mixin OptionsMixin options
+    ) {
+        def userManager = createApplicationContext(options).getBean(UserManager)
+        if (userManager.isUsingMatrixBasedPermissions()) {
```
schnatterer, line 30: Why does this logic belong to the CLI and not to the `UserManager`?

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
```diff
@@ -0,0 +1,51 @@
...
+            @CommandLine.Mixin OptionsMixin options
+    ) {
+        def userManager = createApplicationContext(options).getBean(UserManager)
+        userManager.createUser(username, password)
+    }
+
+    @CommandLine.Command(name = "set-permission", description = "adds user to jenkins", mixinStandardHelpOptions = true)
+    void grantPermission(
+            @CommandLine.Parameters(paramLabel = "username", description = "The username to grant the permission to") String username,
+            @CommandLine.Parameters(paramLabel = "permission", description = "The permission to grant") String permission,
```
schnatterer, line 26: Normally, I'd have said that the description of the `permission` param should contain the possible values. But since we'll plan to throw away the sub command once we're migrated the Jenkins logic from bash to groovy, this doesn't matter. Am I correct?

src/main/groovy/com/cloudogu/gitops/dependencyinjection/JenkinsFactory.groovy
```diff
@@ -0,0 +1,36 @@
...
+        return new ApiClient(
+                config.jenkinsUrl,
+                config.jenkinsUsername,
+                config.jenkinsPassword,
+                client
+        )
+    }
+
+    @Singleton
+    OkHttpClient okHttpClient() {
```
schnatterer, line 30: It's likely that we'll use `OkHttpClient` for other use cases then Jenkins in the future. Would simply re-use this instance and move it to a more generic Factory then, or is this client somewhat specific to Jenkins so that we should name or type it right now?

src/main/groovy/com/cloudogu/gitops/dependencyinjection/JenkinsFactory.groovy
```diff
@@ -0,0 +1,36 @@
...
+import com.cloudogu.gitops.utils.InMemoryCookieJar
+import jakarta.inject.Singleton
+import okhttp3.OkHttpClient
+import io.micronaut.context.annotation.Factory
+
+@Factory
+class JenkinsFactory {
+    private JenkinsCli.OptionsMixin config
+
+    JenkinsFactory(JenkinsCli.OptionsMixin config) {
```
schnatterer, line 15: I'd like to discuss the usage of CLI code in the application

Until now we kept User Interface (CLI) separate from application/core like the general idea of onion architecture. See  `ApplicationConfigurator`.
We never made that decision explicit.  Let's discuss it and then decide how to proceed.

src/main/groovy/com/cloudogu/gitops/jenkins/ApiClient.groovy
```diff
@@ -0,0 +1,66 @@
+package com.cloudogu.gitops.jenkins
+
+import groovy.json.JsonSlurper
+import okhttp3.Credentials
+import okhttp3.FormBody
+import okhttp3.OkHttpClient
+import okhttp3.Request
+
+class ApiClient {
```
schnatterer, line 9: I'd like to have some log statements here.
E.g. trace logs, on protocol-level.
I can imagine it would be interesting to see the script that fails in the logs. Or when something fails regarding the Crumb.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
```diff
@@ -0,0 +1,78 @@
...
+class UserManager {
+    private ApiClient apiClient
+
+    UserManager(ApiClient apiClient) {
+        this.apiClient = apiClient
+    }
+
+    void createUser(String username, String password) {
+        def result = apiClient.runScript("""
+            def realm = Jenkins.getInstance().getSecurityRealm()
```
schnatterer, line 17: Would it be possible to load the Groovy Scripts from files? That way they would be treated as proper code, not just some arbitrary string. Makes it easier to develop them as well.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    void grantPermission(String username, Permissions permission) {
+        def result = apiClient.runScript("""
+            import org.jenkinsci.plugins.matrixauth.PermissionEntry
+            import org.jenkinsci.plugins.matrixauth.AuthorizationType
+
+            def permissions = Jenkins.getInstance().getAuthorizationStrategy().getGrantedPermissionEntries()
+            permissions.computeIfAbsent(jenkins.metrics.api.Metrics.VIEW) {
```
schnatterer, line 36: Why are we doing this? This looks like specific to one specific `Permission` that we enumerated and is independent of the param `permission`.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
```diff
@@ -0,0 +1,78 @@
...
+              new HashSet<>()
+            }
+            print(permissions[$permission].add(new PermissionEntry(AuthorizationType.USER, '${escapeString(username)}')))
+        """)
+
+        log.debug("Granted user $username permission $permission")
```
schnatterer, line 42: We have debug logs 🥳 
I always put them *before*  something happens. So when something fails, I can see where it belonged to. A success message is also good, of course. Do we want both? Can we have some kind of AOP-like options in Micronaut  that allows us to do them declaratively?🙈

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
```diff
@@ -0,0 +1,78 @@
...
+
+        if (result !in ["true", "false"]) { // true == was already in set, false == was not already in set
```
schnatterer, line 44: I don't understand this line, even though it's commented

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
```diff
@@ -0,0 +1,51 @@
...
+            @CommandLine.Mixin OptionsMixin options
+    ) {
+        def userManager = createApplicationContext(options).getBean(UserManager)
+        if (userManager.isUsingMatrixBasedPermissions()) {
+            userManager.grantPermission(username, UserManager.Permissions.valueOf(permission))
```
schnatterer, line 31: Wording: why do we "grant" permissions in groovy and "set" permissions on the CLI side? If there aren't any strong reasons,  I'd prefer a single term

src/main/groovy/com/cloudogu/gitops/utils/InMemoryCookieJar.groovy
```diff
@@ -0,0 +1,29 @@
+package com.cloudogu.gitops.utils
+
+import okhttp3.Cookie
+import okhttp3.CookieJar
+import okhttp3.HttpUrl
+import org.jetbrains.annotations.NotNull
+
+import java.util.function.Predicate
+
+class InMemoryCookieJar implements CookieJar {
```
schnatterer, line 10: No tests for this class. Ratio of cost and benefit not good enough?

schnatterer (changes requested):
There are a couple of things in this PR that stand out positively:

* the Jenkins groovy shell will give us much more options and likely more stability than continuing to POST forms to Jenkins
* the Jenkins sub-command allows as to gradually migrate from shell to groovy while not having to build two static images
* the bash function `createUser` , makes it easy to test it e2e and easier to use in the next step
* HTTP retries with back-off 🥳
* good catch that we didn't use the micronaut-http-client at all 🙈
* the test using mock webserver

I mentioned annotated some things worth changing or discussion  in the code

scripts/jenkins/init-jenkins.sh
```diff
@@ -73,6 +73,10 @@ function waitForJenkins() {
   echo ""
 }
 
+function createUser() {
```
schnatterer, line 76: We might want to pass `-d` or  `-x` here or otherwise we won't see any of those debug or trace logs from groovy.
Once we've done that the Jenkins subcommand will complain about 
`Unknown option: '-x' `, for example. The option is actually evaluated in `GitopsPlaygroundCliMain` so it actually works. One solution would be to add `-x` and `-d` to the `jenkins` subcommand.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
```diff
@@ -0,0 +1,78 @@
...
+            throw new RuntimeException("Failed to add permission $permission to $username: $result")
+        }
+    }
+
+    boolean isUsingMatrixBasedPermissions() {
+        def result = apiClient.runScript("print(Jenkins.getInstance().getAuthorizationStrategy().class)")
+
+        return result == "class hudson.security.GlobalMatrixAuthorizationStrategy"
```
schnatterer, line 52: I tried this with our BCE-Instance and it has `ProjectMatrixAuthorizationStrategy`, so I can set the permissions there. Should we allow both? Would `grantPermission()` also work?

scripts/utils.sh
CloudoguSiebels: In theory, having secrets as command line arguments is bad practice anyway. You can easily leak them in logging, but more importantly you always leak them in `proc/$pid/cmdline` in a multi-user environment. When moving to config files and/or environment, we could simply log the entire command line here. However, that seems to be out of reach right now.

We could move logging the entrypoint into groovy where we have more information on what to log and what not to log. What do you think?

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
CloudoguSiebels: No reason really. IDE didn't suggest it, and I didn't think about it. I agree though.

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
CloudoguSiebels: Yes, I want to migrate the entire Jenkins logic away from this subcommand. However, listing possible values is easy enough with `${COMPLETION-CANDIDATES}`.

src/main/groovy/com/cloudogu/gitops/dependencyinjection/JenkinsFactory.groovy
CloudoguSiebels: I envision to reuse the http client. Therefore, I created a separate Factory for it.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
CloudoguSiebels: We could. I decided against because

* IDEA hat support for inline code in strings
* We need to replace strings within the code, e.g. username and password when creating a user.

The latter could be solved by convention that groovy-scripts define a function that can be invoked and we concatenate the function definition and function call in the `UserManager`. 
That is complexity I didn't deem worthy right now. Do you think otherwise?

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
CloudoguSiebels: I missed to replace the concrete permission with $permission.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
CloudoguSiebels: Did my change clarify things?

In general, it is difficult to recognize script errors within the jenkins script. Therefore, I want to compare the result to known good values, and report an error on everything else.

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
CloudoguSiebels: Yes, I was thinking about this as well. Having a second opinion going in the same direction justifies a change.

src/main/groovy/com/cloudogu/gitops/utils/InMemoryCookieJar.groovy
CloudoguSiebels: I managed to replace this class by a java.net-adapter.

scripts/jenkins/init-jenkins.sh
CloudoguSiebels: We evaluate the option in two places: CliMain for picocli's own logging. And `ApplicationConfigurator` for our own logging.

I don't think that tracing needs be part of the configuration as that is handled in the infrastructure layer. As a result, I moved it out of there into the parent cli class to handle the logging for all of them.

src/main/groovy/com/cloudogu/gitops/cli/JenkinsCli.groovy
schnatterer: Completion candidates is a good idea. I only wonder if they will work when `paramters` is a string and not an enum at CLI-level.

But I'd leave it at that, because the CLI hopefully won't be used by end-users.

scripts/utils.sh
schnatterer: In general, logging in groovy would be a good idea.
As the different entrypoints are hopefully only temporary, I'd say we leave it as is for now.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
schnatterer: Yeah, you're right. At this point this is probably too much effort :+1:

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
schnatterer: Your addition to the comment clear up my prior confusion :smile: :+1:

scripts/jenkins/init-jenkins.sh
schnatterer: That's an interesting insight and a smart implementation :+1:

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
schnatterer: :bulb: TIL that idea can inject language support via Alt + Enter and "Inject language. That'd be worth a comment. But where? Maybe at `ApiClient.runScript()`?
I can imagine that GString interpolation inside a groovy file can be complicated and loading external files makes thing more complicated in general. Let's start as is.

src/main/groovy/com/cloudogu/gitops/dependencyinjection/JenkinsFactory.groovy
schnatterer: That's the only remaining issue. What are your thoughts on this @CloudoguSiebels?

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
CloudoguSiebels: In PhpStorm you can mark the injected code like this: 

https://www.jetbrains.com/help/phpstorm/using-language-injections.html#use-lang-annotation

I didn't find any suggestion for that using alt+enter. We could try it the same was PhpStorm does it.

src/main/groovy/com/cloudogu/gitops/dependencyinjection/JenkinsFactory.groovy
CloudoguSiebels: Oh, I missed to reply, sorry.

I think that DI belongs to the infrastructure layer. Therefore I think that depending on a other infrastructure code such as Picocli is okay. 

That does not address you concern regarding cli vs non-cli. One could argue that the mixin is independent from picocli as picocli only interfaces using annotations. However, I do think that this is not the appropriate solution.

I would suggest to implement a layer of abstraction regarding the configuration similar to ApplicationConfigurator. I try to change this in the evening; or you can change it yourself if you feel like it.

src/main/groovy/com/cloudogu/gitops/jenkins/UserManager.groovy
schnatterer: Good point. :bulb: Reading the doc helps :see_no_evil: 
https://www.jetbrains.com/help/idea/using-language-injections.html#use-language-injection-comments
I'll add a `/* language=groovy */ ` instead of a comment!

schnatterer: The devil in the detail :roll_eyes: 
IDEA-lang-injection

schnatterer:
This PR is one huge step towards getting out of our Jenkins-related Bash-hell :partying_face:

PR #117, opened 2023-06-23
Make images configurable

CloudoguSiebels:

PR #121, opened 2023-07-05
Upgrade SCMM

schnatterer:
Avoids errors caused by accidental breaking change in scmm-review-plugin with gitops-build-lib.

The problem was fixed in 2.27.1, which requires SCM-Manager 2.44.0.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ properties([
     ])
 ])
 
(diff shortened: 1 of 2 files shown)


schnatterer:
The problem reported in #117 was fixed here: 
scm-manager/scm-review-plugin@3e643a8

So we should be able to merge

PR #122, opened 2023-07-07
Add parameter to prefix namespaces

CloudoguSiebels:
When deploying into an existing SCM-Manager/Jenkins and/or Kubernetes cluster, we want to avoid conflicting names.

We introduce the parameter `--name-prefix` which adds a prefix to the following:

* Kubernetes namespaces
* SCM-Manager namespaces
* Jenkins job

We also renamed the namespaces `argocd-staging` and `argocd-production` to `<prefix>-example-apps-staging` and `-production` to better reflect their purpose: deploying the example apps.


applications/argocd/nginx/helm-jenkins/Jenkinsfile.tpl
```diff
@@ -53,13 +54,14 @@ node('docker') {
...
+</#noparse>
+                    k8sVersion : env.${namePrefixForEnvVars}K8S_VERSION,
                     buildImages          : [
-                            helm: 'ghcr.io/cloudogu/helm:3.10.3-1',
-                            kubectl: 'lachlanevenson/k8s-kubectl:v1.25.4',
-                            kubeval: 'ghcr.io/cloudogu/helm:3.10.3-1',
-                            helmKubeval: 'ghcr.io/cloudogu/helm:3.10.3-1',
-                            yamllint: 'cytopia/yamllint:1.25-0.7'
+                            helm: '<#if images.helm != "">${images.helm}<#else>ghcr.io/cloudogu/helm:3.10.3-1</#if>',
+                            kubectl: '<#if images.kubectl != "">${images.kubectl}<#else>lachlanevenson/k8s-kubectl:v1.25.4</#if>',
```
schnatterer, line 61: That's a fine example of boy scout rule!
We could even go one step further and rely on the defaults in `ApplicationConfigurator`, effectively getting rid of a lot of redundant `<#if ..><#else>` block in several `Jenkinsfile`s. Or am I missing something?

docs/architecture-decision-records.md
```diff
@@ -3,6 +3,55 @@ Architecture Decision Records
...
+* Micronaut-compatible Library
+
+### Decision Outcome
+
+#### Groovy Templating
+
+Groovy Templating has a very small footprint and does not rely on a third-party library.
+However, it relies on dynamic code execution and is therefore incompatible with GraalVM.
+
+#### Micronaut-compatible Library
```
schnatterer, line 36: Where is the connection to micronaut regarding those libs? Are they listed somewhere as being officially compatible?

applications/argocd/nginx/helm-jenkins/Jenkinsfile.tpl
CloudoguSiebels: Good catch! Absolutely!

CloudoguSiebels:
I am quite happy that I still get code highlighting for YAML files. In most cases the templates are valid YAML when we are only using variables and not directives.

replacing `.tpl` with `.ftl` sounds good to me.

docs/architecture-decision-records.md
schnatterer: Yes, definitely a good idea. When reading the ADR the connection to Micronaut wasn't perfectly obvious to me. The link would be a good addition.

PR #125, opened 2023-07-13
Use dependency injection

CloudoguSiebels:
Using dependency injection allows us to be very explicit about
dependencies without needing to manually wire them with every change.
Explicitness helps us write tests, as it is obvious which dependencies
need to be mocked/changed/injected without risking to run code on a
real cluster.

Furthermore, by using `@Singletons` we gain the advantage to actually
implement services which use memoization.
For instance, the ScmmRepos don't need to be cloned as much, if we
utilize this.


src/main/groovy/com/cloudogu/gitops/features/argocd/ArgoCD.groovy
```diff
@@ -26,65 +37,66 @@ class ArgoCD extends Feature {
...
     private List<RepoInitializationAction> gitRepos = []
 
     private String password
     
-    protected File argocdRepoTmpDir
-    private RepoInitializationAction argocdRepoInitializationAction
-    protected File clusterResourcesTmpDir
-    protected File exampleAppsTmpDir
-    protected File nginxHelmJenkinsTmpDir
+    protected RepoInitializationAction argocdRepoInitializationAction
```
CloudoguSiebels, line 45: Most of the changes in this class are used to replace `tmpDir` with `RepoInitializationAction` to remove usage of the deprecated constructor of the ScmmRepo.

You can review these changes separately to dependency injection by looking at the two different commits.

PR #145, opened 2023-08-23
 Retry on 403/401 for creating jenkins properties

CloudoguSiebels:
We observed temporary 403/401s when creating these properties directly
after restarting jenkins.
We simply retry these requests and hope that they will not fail
permanently.


schnatterer:
LGTM

schnatterer (approved):

PR #148, opened 2023-09-05
Use JGit rather than git binary

CloudoguSiebels:


src/main/groovy/com/cloudogu/gitops/scmm/jgit/InsecureCredentialProvider.groovy
```diff
@@ -0,0 +1,38 @@
+package com.cloudogu.gitops.scmm.jgit
+
+import org.eclipse.jgit.errors.UnsupportedCredentialItem
+import org.eclipse.jgit.transport.CredentialItem
+import org.eclipse.jgit.transport.CredentialsProvider
+import org.eclipse.jgit.transport.URIish
+
+class InsecureCredentialProvider extends CredentialsProvider {
```
schnatterer, line 8: Could you please add some comment here? The checks for prompts look a bit awkward though I'm sure you thoroughly checked before that there is no easier option.  An answer to the question "Why was this implemented that way?" would surely be helpful for future reference.

schnatterer (changes requested):
@CloudoguSiebels 
This looks very good. I'm so glad to get rid of the awful credential handling in the code and the shell-out.

The only questions I have is about the insecure handling.
The former code did not implement the insecure flag, did it? Was this a learning from errors occurred recently?

CloudoguSiebels:
> The former code did not implement the insecure flag, did it? Was this a learning from errors occurred recently?

Yes it did by using an environment variable. 

https://github.com/cloudogu/gitops-playground/blob/main/scripts/apply.sh#L71

*This* wasn't failing recently. However, recent failures prompted me to test this case :)

src/main/groovy/com/cloudogu/gitops/scmm/jgit/InsecureCredentialProvider.groovy
CloudoguSiebels: Yep, will do. 

According to threads in Stackoverflow, JGit was developed with a interactive UI in mind.

schnatterer: :bulb: I think it's helpful to add a link to SO in the comments.

schnatterer (approved):

PR #150, opened 2023-09-13
Add option to configure playground by file

CloudoguSiebels:


src/main/groovy/com/cloudogu/gitops/cli/GenerateJsonSchema.groovy
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode
+import io.micronaut.context.ApplicationContext
+
+/**
+ * Generates the JSON Schema for the configuration file and prints it to stdout.
+ * We save the schema to docs/configuration.schema.json. JsonSchemaGeneratorTest ensures that this is updated.
+ *
+ * @see com.cloudogu.gitops.config.schema.Schema
+ */
+class GenerateJsonSchema {
```
schnatterer, line 13: Wouldn't integration into the normal CLI make the whole process simpler?

We wouldn't need to store the Schema in the repo.
Users wouldn't have to search for the right version of the schema but could just generate the appropriate one with the version they use anyway.

pom.xml
```diff
@@ -247,6 +247,24 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
schnatterer, line 250: ADR for the libs would be interesting. Doesn't Jackson offer something regarding to Schema?

src/main/groovy/com/cloudogu/gitops/config/schema/JsonSchemaGenerator.groovy
```diff
@@ -0,0 +1,21 @@
...
+import com.github.victools.jsonschema.generator.OptionPreset
+import com.github.victools.jsonschema.generator.SchemaGenerator
+import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder
+import com.github.victools.jsonschema.generator.SchemaVersion
+import jakarta.inject.Singleton
+
+@Singleton
+class JsonSchemaGenerator {
+    ObjectNode createSchema() {
+        SchemaGeneratorConfigBuilder configBuilder = new SchemaGeneratorConfigBuilder(SchemaVersion.DRAFT_2019_09, OptionPreset.PLAIN_JSON)
```
schnatterer, line 15: Why this exact version?
Is it the best common denominator between generator and validator?

We could think about storing the string behind the Schema Version in a constant and use it in both the generator and the validator to clearly express that these values have to be the same. WDYT?

src/main/groovy/com/cloudogu/gitops/cli/GenerateJsonSchema.groovy
CloudoguSiebels: But I want to use it in my IDE. It is much simpler to reference https://raw.githubusercontent.com/cloudogu/gitops-playground/main/docs/configuration.schema.json than to generate it ourselves every time. After all this is a process designed for the users, not the developers of the playground.

When the schema changes, a unit test will fail and the developer is immediately prompted to do something. In that case the test also outputs what to do to resolve the error. This is also mentioned in the docblock for GenerateJsonSchema.groovy.

src/main/groovy/com/cloudogu/gitops/config/schema/JsonSchemaGenerator.groovy
CloudoguSiebels: We could update to the 2020-12 version. Both generator and validator support this.

`SchemaVersion` is an enum. When saving the string in a separate place, we would need to create the enum from string which is smelly.
We could try to infer the schema version by the `$schema` key within the json schema file. The library might offer something for that.

pom.xml
CloudoguSiebels: This library is officially linked at json-schema.org https://json-schema.org/implementations.html#schema-generators

Jackson supports this only via a third party module. This didn't seem as stable though https://github.com/FasterXML/jackson-module-jsonSchema#future-plans-lack-thereof

src/main/groovy/com/cloudogu/gitops/config/schema/JsonSchemaGenerator.groovy
CloudoguSiebels: The `SpecVersionDetector` is able to infer the schema version.

CloudoguSiebels:
Yes, you are correct. We need them to be null by default.

schnatterer:
And one last thing: In the future when changing something on either the CLI or the Schema/config file we should probably consider doing the same change on the both.
How can we make that sure? A comment on both the Schema and the CLI class?

CloudoguSiebels:
> Wouldn't it be more uniform to either use Json in all names or in none?

I think these operate on different levels. `JsonSchemaGenerator` and `GenerateJsonSchema` operate on json-schema by specification.
`Schema` is not coupled to json-schema, rather is a generic schema for our config object. In future work, this could be our designated config object, dropping schema from the name then.
`SchemaValidator` however is json-schema specific.


> And one last thing: In the future when changing something on either the CLI or the Schema/config file we should probably consider doing the same change on the both.

Yes. I don't have a good solution so that you don't have to think about it. A comment is the best I can come up with, too.

README.md
```diff
@@ -205,6 +205,15 @@ You can get a full list of all options like so:
...
 
+##### Configuration file
+
+You can also use a configuration file to specify the parameters (`--config-file`).
+That file must be a YAML file. You can find the schema [here](https://raw.githubusercontent.com/cloudogu/gitops-playground/main/docs/configuration.schema.json).
+
+Note that currently, only part of the configuration parameters are supported.
+
+See [here](https://www.jetbrains.com/help/ruby/yaml.html#remote_json) how to configure IntelliJ IDEA to use the schema and offer autocompletion and validation.
+
```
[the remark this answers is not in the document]
CloudoguSiebels: > How did you test this? Using docker run -v?

Yes, exactly.

> And what about precedence? When both params and config-file are present, who wins?

As described in the ticket, cli overwrites config file. Additionally, there is a test for it.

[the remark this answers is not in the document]
CloudoguSiebels: > implement conventions, e.g. load from a key called config.yaml and use the current namespace of the kubeconfig.

I would be fine with that. What do you think?

src/main/groovy/com/cloudogu/gitops/cli/GenerateJsonSchema.groovy
CloudoguSiebels: > Still, wouldn't it make sense to provide an option for refactoring this class to be part of the proper CLI instead of a separate entry point?

Yes. However, I was hesitant to do so due to the number of parameters that don't make much sense for the schema. Furthermore, we would need to implement the switch within bash as well.

CloudoguSiebels: > Talking about users: from a user's point of view an example YAML with all options would also help me.

Yes, I thought about this as well. First, I printed the default `ApplicationConfigurator` value. However, these include more fields than appropriate. Maybe we can parse the command line options and output yaml based on the schema. Similar to `kubectl create ns foobar --dry-run=client -oyaml`

README.md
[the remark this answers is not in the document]
schnatterer: That'll be a pragmatic first solution that allows for adding more params later, if we find it necessary :+1:

src/main/groovy/com/cloudogu/gitops/cli/GenerateJsonSchema.groovy
schnatterer: > Yes. However, I was hesitant to do so due to the number of parameters that don't make much sense for the schema. Furthermore, we would need to implement the switch within bash as well.

Wouldn't a subcommand avoid the params?
e.g. ` schema`  or `config schema` (if we decided to use`config example` or similar)?
Yes, we would need it in bash, but I think this could be a simple if statement at the beginning that exits after `runGroovy`.

For the example, couldn't we just convert the Schema class to YAML?

But maybe the whole discussion is "premature optimization".
If it turns out to be complicated I'd leave that for the future.

CloudoguSiebels: > A kind of dry-run wouldn't meet my requirement to easily see all options. For example when using a helm chart, looking at the default values.yaml makes it easy for me to use.

We have either passed parameters or default values. I would output the parsed and merged config, containing all values.

```
docker run [...] --argocd --skip-helm-update  --output-config-file
[...]
---
images:
  kubectl: "lachlanevenson/k8s-kubectl:v1.25.4"
  helm: "ghcr.io/cloudogu/helm:3.10.3-1"
  kubeval: "ghcr.io/cloudogu/helm:3.10.3-1"
  helmKubeval: "ghcr.io/cloudogu/helm:3.10.3-1"
  yamllint: "cytopia/yamllint:1.25-0.7"
  nginx: null
features:
  argocd:
    url: ""
  mail:
    url: ""
  monitoring:
    active: false
    grafanaUrl: ""
    helm:
      grafanaImage: ""
      grafanaSidecarImage: ""
      prometheusImage: ""
      prometheusOperatorImage: ""
      prometheusConfigReloaderImage: ""
  secrets:
    vault:
      mode: ""
      url: ""
      helm:
        image: ""
    externalSecrets:
      helm:
        image: ""
        certControllerImage: ""
        webhookImage: ""
  exampleApps:
    petclinic:
      baseDomain: ""
    nginx:
      baseDomain: ""

```

schnatterer: :bulb: That looks good!

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -180,9 +183,20 @@ class GitopsPlaygroundCli  implements Runnable {
...
+            System.exit(1)
+        }
+
+        def appContext = ApplicationContext.run()
+        ApplicationConfigurator applicationConfigurator = appContext.getBean(ApplicationConfigurator)
         if (configFile) {
             applicationConfigurator.setConfig(new File(configFile))
+        } else if (configMap) {
+            def k8sClient = appContext.getBean(K8sClient)
+            def configValues = k8sClient.getConfigMap(configMap, 'config.yaml')
```
schnatterer, line 202: Here some simple error handling would be good, e.g.
if empty, print that the key `config.yaml` could not be found in configMap 

And if you still can bear with me a test would be nice :see_no_evil: 

I'll stop having more ideas for this PR now, promise :)

PR #151, opened 2023-09-15
Retry HTTP request when socket times out

CloudoguSiebels:
We observed socket timeouts during jenkins restarts, especially when the host has high load. As a countermeasure, we retry these cases.


schnatterer:
LGTM!
Very nice test :+1:

PR #153, opened 2023-09-20
Improved retrying for invalid jenkins crumbs

CloudoguSiebels:
We were running into an issue where jenkins was restarted after we fetched a crumb but before we used that crumb.

1. Fetch crumb
2. Restart Jenkins
3. Use crumb e.g. in /scriptText

Due to the restart, the crumb is invalid (either due to the crumb itself or the associated session).
The previous handling would then retry the /scriptText request without refreshing the crumb.

We changed the behaviour in the following responsibilities:

1. RetryInterceptor is responsible for retrying HTTP level issues (e.g. status codes or timeouts)
2. ApiClient is responsible for application (aka Jenkins) level issues (i.e. HTTP status code 401 or 403 during restarts or invalid crumbs)

As a result, we don't leak special cases for Jenkins into the RetryInterceptor anymore and it can be used for SCMM APIs as well.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/dependencyinjection/RetrofitFactory.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/dependencyinjection/RetrofitFactory.groovy
@@ -1,6 +1,7 @@
 package com.cloudogu.gitops.dependencyinjection
 
 import com.cloudogu.gitops.config.Configuration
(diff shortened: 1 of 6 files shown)


schnatterer:
Really nice solution using the closure and great to have proper tests for this rather complex situation.

PR #155, opened 2023-09-22
Stop execution when a step fails

### LINKED ISSUE #7: Spinner prints "ok" on failure
The new `apply.sh` looks really awesome!

One downside: when one of the initializations fails, it still prints `ok` and continues the initialization.

I'd prefer if it printed `failed` and stops the execution.

You can reproduce this by making an `init` function `return 1` for example.

CloudoguSiebels:
rather than continuing with the next step.
This only affects when not running with -x

closes #7

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -44,7 +44,13 @@ function spinner() {
         done
         printf $reset
     done
-    echo " [ok] $info"
+
+    if wait -n $pid; then
+      echo " [ok] $info"
+    else
+      echo " [failed] $info"
+      exit 1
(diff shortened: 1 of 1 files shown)


CloudoguSiebels:
@schnatterer tried a similar solution using wait in the past (see #7). I could not find any issues with it though.

schnatterer:
@CloudoguSiebels THB I can't remember the exact circumstances but tested your change and it looks as if it does the job. At last, our bash error handling no longer provides unpleasant surprises for our users :clap:

PR #170, opened 2023-12-17
added feature configure external Mailserver, corrected settings for g…

movray:
…rafana contact point when internal Mailserver is used, added pependency check if internal or externa Mailserver is used


scripts/apply.sh
```diff
@@ -439,8 +439,7 @@ function printUsage() {
...
-    -- "$@")
+    --long help,config-file:,config-map:,output-config-file,destroy,argocd,argocd-url:,debug,remote,base-url:,username:,password:,jenkins-url:,jenkins-username:,jenkins-password:,jenkins-metrics-username:,jenkins-metrics-password:,registry-url:,registry-path:,registry-username:,registry-password:,internal-registry-port:,scmm-url:,scmm-username:,scmm-password:,kubectl-image:,helm-image:,kubeval-image:,helmkubeval-image:,yamllint-image:,grafana-url:,grafana-image:,grafana-sidecar-image:,prometheus-image:,prometheus-operator-image:,prometheus-config-reloader-image:,external-secrets-image:,external-secrets-certcontroller-image:,external-secrets-webhook-image:,vault-url:,vault-image:,nginx-image:,trace,insecure,yes,skip-helm-update,metrics,monitoring,grafana-email-from,grafana-email-to,argocd-email-from,argocd-email-to-admin,argocd-email-to-user,mail,mailhog-url:,smtp-address,smtp-port,smtp-user,smtp-password,vault:,petclinic-base-domain:,nginx-base-domain:,name-prefix: \    -- "$@")
```
schnatterer, line 442: Please add colons after all email-related params.

This way the user can choose to pass the params with or without equals sign.

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -117,11 +119,37 @@ class GitopsPlaygroundCli  implements Runnable {
     @Option(names = ['--vault-url'], description = 'Sets url for vault ui')
     private String vaultUrl
 
-    // args group mail
-    @Option(names = ['--mail'], description = 'Installs MailHog as Mail server.', scope = CommandLine.ScopeType.INHERIT)
-    private Boolean mail
-    @Option(names = ['--mailhog-url'], description = 'Sets url for mailhog')
-    private String mailhogUrl
+    @ArgGroup(exclusive = true)
+    MailHogUrlIfNotBaseUrl mailHogUrlIfNotBaseUrl
```
schnatterer, line 123: IMO we don't need this. It's OK to set both mailhog-url and base-url.
With base-url you set defaults for all ingresses. If you want to have a custom ingress only for mailho, you can also set mailhog-url,

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
```diff
@@ -185,6 +186,8 @@ class ApplicationConfigurator {
             newConfig.registry["internal"] = false
         if (newConfig['features']['secrets']['vault']['mode'])
             newConfig['features']['secrets']['active'] = true
+//        if (newConfig.['features']['mail']['mailhog'] || smt-paramer)
```
schnatterer, line 189: Why didn't you set mail.active here:
config.features['mail']['mailhog'] || config.features['mail']['externalMailserver'] ? true : false

This way you could have left the mail.active in ArgoCD and Prometheus as they were.

src/main/groovy/com/cloudogu/gitops/features/argocd/ArgoCD.groovy
```diff
@@ -312,7 +312,11 @@ class ArgoCD extends Feature {
                             ]
                     ],
                     mail: [
-                            active: config.features['mail']['active']
+                            active: config.features['mail']['mailhog'],
```
schnatterer, line 315: This way we will not get any `notifications.argoproj.io/subscribe.email` annotation when we use the external mail server. Please use mail.active here.

schnatterer:
Please add new params also to Schema, so they can be configured via config file

src/main/groovy/com/cloudogu/gitops/features/argocd/ArgoCD.groovy
movray: changed

applications/cluster-resources/monitoring/prometheus-stack-helm-values.ftl.yaml
```diff
@@ -103,10 +103,41 @@ grafana:
...
+          is_default: true
+          receiver: email
+          routes:
+          - receiver: email
+          group_by: ["grafana_folder", "alertname"]
+ <#if mail.externalMailserver?has_content>
+  env:
+    GF_SMTP_ENABLED: true
+    GF_SMTP_FROM_ADDRESS: ${monitoring.grafanaEmailFrom}
+    GF_SMTP_HOST: ${mail.externalMailserver}:${mail.externalMailserverPort}
```
schnatterer, line 132: What happens if port is not set? Shouldn't you check if it is set here?

You should probably also only set Env Vars for user and password if  externalMailserverUser and externalMailserverPassword are set.

Same goes for Argo CD.

Please write tests.

applications/cluster-resources/monitoring/prometheus-stack-helm-values.ftl.yaml
```diff
@@ -103,10 +103,41 @@ grafana:
...
+    notification-policies.yaml:
+      apiVersion: 1
+      policies:
+        - orgId: 1
+          is_default: true
+          receiver: email
+          routes:
+          - receiver: email
+          group_by: ["grafana_folder", "alertname"]
+ <#if mail.externalMailserver?has_content>
```
schnatterer, line 128: Why do you use `has_ _content` here and not `??

argocd/argocd/argocd/values.ftl.yaml
```diff
@@ -95,14 +95,21 @@ argo-cd:
...
 
   notifications:
-    enabled: ${mail.active?string('true', 'false')}
+    enabled: <#if mail.active??>true<#else>false</#if>
     argocdUrl: https://localhost:9092
-<#if mail.active == true>
+<#if mail.active??>
     notifiers:
       service.email: |
+      <#if mail.externalMailserver?has_content>
```
schnatterer, line 103: See prometheus stack

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -274,8 +285,12 @@ class GitopsPlaygroundCli  implements Runnable {
                                 emailToAdmin : emailToAdmin
                         ],
                         mail: [
-                                active    : mail,
+                                mailhog: mailhog,
                                 url       : mailhogUrl,
+                                externalMailserver : smtpAddress,
```
schnatterer, line 289: Using two different names for the same thing makes things more complex than necessary.

We as developers will know have to remember that
`externalMailserver` is synonym to `smtpAddress` and the same goes for port, user and password. 

And this causes confusion! You can already see it in Schema.groovy (see bellow).
We should use the same terms for the same things.

This would either mean using `smtp*` internally or `external-mail-server-*` on the CLI. I don't care which, prefer the the less time consuming one :grimacing:

src/main/groovy/com/cloudogu/gitops/config/schema/Schema.groovy
```diff
@@ -78,8 +78,12 @@ class Schema {
     }
 
      static class MailSchema {
-         boolean active = true
-         String url = ""
+         boolean mailhog = true
```
schnatterer: In order for this to work the hierarchy of the objects in this class has to match the hierarchy of the `config` map.

For example:
`config` map: contains
* features.mail.url
* features.mail.externalMailserver

In `Schema.groovy` we have: 
* features.mail.mailhogUrl
* features.mail.smtpAddress

This leads to the values from config file to being ignored.

I know, the solution here is somewhat complex and we had errors like this before.

So I added a test: `ApplicationConfiguratorTest.config file has only fields that are present in default values` that fails. Please fix it.

schnatterer (changes requested):
We made some progress!
I still see some issues here, though. Sorry, @movray .

argocd/argocd/argocd/values.ftl.yaml
schnatterer: `has_content` checks for not null and empty string, `??` not null and `true` for booleans.
In our case `has_content` is probably more reliable, because we often use empty string as default value.

applications/cluster-resources/monitoring/prometheus-stack-helm-values.ftl.yaml
schnatterer: has_content checks for not null and empty string, ?? not null and true for booleans.
In our case has_content is probably more reliable, because we often use empty string as default value.

schnatterer (approved):

schnatterer:
I tested this

```
scripts/init-cluster.sh --bind-ingress-port=80  && docker run --rm -u $(id -u) \                                     
    -v ~/.config/k3d/kubeconfig-gitops-playground.yaml:/home/.kube/config \
    --net=host gitops-playground --yes --argocd --base-url=http://localhost  --monitoring --vault=dev --mailhog --mailhog-url http://hog.localhost -x  --argocd-email-from=from@argo.cd --argocd-email-to-admin=to-admin@argo.cd \
  --argocd-email-to-user=to-user@argo.cd --grafana-email-from=from@grafa.na --grafana-email-to=to@grafa.na
```

and  this

``` 
 scripts/init-cluster.sh --bind-ingress-port=80  \
&& docker run --rm -u $(id -u) \
    -v ~/.config/k3d/kubeconfig-gitops-playground.yaml:/home/.kube/config \
       -v $(pwd)/gitops-playground.yaml:/config/gitops-playground.yaml \
    --net=host gitops-playground --yes --argocd --base-url=http://localhost  --monitoring --vault=dev --mailhog --mailhog-url http://hog.localhost -x --config-file=/config/gitops-playground.yaml
``` 

with this config file

```yaml
features:   
  mail:  
    mailhog: true # should be ignored  
    smtpAddress: 'smtp.gmail.com'
    smtpPort: 465
    smtpUser: USER
    smtpPassword: APP_PW
  argocd:  
    emailFrom: USER
    emailToAdmin: other email
    emailToUser: other email
  monitoring:  
    grafanaEmailFrom: USER
    grafanaEmailTo: other email
```

And everything seems to work.

There is one thing we should improve, though: The `smtpPassword` is stored in plain text in git 😱!

In argocd we can avoid this by using the [argocd-notifications-secret](
https://github.com/argoproj/argo-cd/blob/v2.9.5/docs/operator-manual/notifications/index.md#getting-started). But what about Grafana?

PR #178, opened 2024-03-11
Add paramters to define the gitAuthor 

yannickchristhomas:
The author for the inital commits can now be defined with the parameters `--git-name` and `--git-email`.


schnatterer:
@yannickchristhomas 
I committed a fix that hopefully fixes the cause for the failing build on this branch. Note that it like is not related to your changes. Presumable it's a concurrency problem, that only occurs every now and then. I probably introduced it myself in 4b7ca871.

schnatterer:
I pondered a while about this, but allowing everyone to write to /root feels wrong, even in a container. I can't provide an exact attack vector but, it would put my security conscience a lot more at peace if we would only allow 
* world read from /root
* and world write to `/root/.config/jgit/config`

So please create the file in `Dockerfile` and make only this file world writable. Everything else looks perfect :+1:

PR #180, opened 2024-03-12
Added new parameter for urlSeparatorHyphen

movray:


schnatterer:
@movray 
Looks almost good to me. Only thing I'm missing is the config file option for `urlSeparatorHyphen`.  Please also have a look if you're happy with the polishing I did.

schnatterer:
For the record: we now have `urlSeparatorHyphen` in the config but it does not work, yet, because we haven't enabled all options after getting rid of bash in #174

PR #187, opened 2024-04-22
Fix gpg import bug for helm

yannickchristhomas:
This change adds the -t 3 flag to the wget command, which means the download attempt will be retried up to three times if connection errors occur.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,7 @@ RUN wget -q -O helm.tar.gz.asc https://github.com/helm/helm/releases/download/v$
 RUN tar -xf helm.tar.gz
 # Without the two spaces the check fails!
 RUN echo "${HELM_CHECKSUM}  helm.tar.gz" | sha256sum -c
-RUN wget -q https://raw.githubusercontent.com/helm/helm/main/KEYS -O- | gpg --import
+RUN wget -t 3 -q https://raw.githubusercontent.com/helm/helm/main/KEYS -O- | gpg --import
 RUN gpg --batch --verify helm.tar.gz.asc helm.tar.gz
 RUN mv linux-amd64/helm /dist/usr/local/bin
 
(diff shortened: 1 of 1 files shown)


schnatterer:
FTR Some context: Our build fails often with this:
```  
18:08:19  ------
18:08:19  Dockerfile:67
18:08:19  --------------------
18:08:19    65 |     # Without the two spaces the check fails!
18:08:19    66 |     RUN echo "${HELM_CHECKSUM}  helm.tar.gz" | sha256sum -c
18:08:19    67 | >>> RUN wget -q https://raw.githubusercontent.com/helm/helm/main/KEYS -O- | gpg --import
18:08:19    68 |     RUN gpg --batch --verify helm.tar.gz.asc helm.tar.gz
18:08:19    69 |     RUN mv linux-amd64/helm /dist/usr/local/bin
18:08:19  --------------------
18:08:19  ERROR: failed to solve: process "/bin/sh -c wget -q https://raw.githubusercontent.com/helm/helm/main/KEYS -O- | gpg --import" did not complete successfully: exit code: 4294967295
``` 

Assumption was that downloading the KEYS file failed. Thus the retries.

I had a closer look at busybox wget, which [seems to ignore](https://git.busybox.net/busybox/tree/networking/wget.c?h=1_36_1&id=1a64f6a20aaf6ea4dbba68bbfa8cc1ab7e5c57c4#n140) the `-t` option altogether. 
"Full" wget (e.g. `apk add wget`) uses [a default of 20](https://www.man7.org/linux/man-pages/man1/wget.1.html) for `-t` .  
But I wonder for what code they retry at all, e.g. this does not retry 🤔
```
wget -O-  https://httpbin.org/status/503
```

We could use ` --retry-on-http-error`, but then we would have to specify a lot of error codes. So I dediced to switch to curl.

PR #219, opened 2024-08-27
Adapt gop graphics

flxebrt:
adaptation of feature and local graphic


schnatterer:
Nice! This makes our graphics more consistent and clearer where the GOP fits in technically.

PR #227, opened 2024-09-13
Prometheus: Implement Network Policies for namespace isolation

nihussmann:
adds the capability to set additional network policies for some use cases (e.g. Openshift Clusters)

### DIFF
--- a/applications/cluster-resources/monitoring/netpols/prometheus-allow-scraping.ftl.yaml
+++ b/applications/cluster-resources/monitoring/netpols/prometheus-allow-scraping.ftl.yaml
@@ -9,7 +9,7 @@ spec:
     - from:
         - namespaceSelector:
             matchLabels:
-              name: ${namePrefix}monitoring
+              kubernetes.io/metadata.name: ${namePrefix}monitoring
           podSelector:
             matchLabels:
               prometheus: kube-prometheus-stack-prometheus
--- a/applications/cluster-resources/monitoring/netpols/prometheus-allow-scraping.ftl.yaml
+++ b/applications/cluster-resources/monitoring/netpols/prometheus-allow-scraping.ftl.yaml
@@ -1,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-prometheus-scraping
+  namespace: ${namespace}
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              name: ${namePrefix}monitoring
+          podSelector:
+            matchLabels:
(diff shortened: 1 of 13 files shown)


src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -138,6 +138,8 @@ class GitopsPlaygroundCli  implements Runnable {
     private Boolean skipCrds
     @Option(names = ['--namespace-isolation'], description = NAMESPACE_ISOLATION_DESCRIPTION)
     private Boolean namespaceIsolation
+    @Option(names = ['--netpols'], description = NETPOLS)
```
schnatterer, line 141: Please change to `NETPOLS_DESCRIPTION`. I have been wondering for some time as well if we actually need the `_DESCRIPTION` suffix, but for now every other property has it, so for the sake of uniformity, I'd keep it.
Or remove all other `_DESCRIPTION` suffixes?

applications/cluster-resources/monitoring/netpols/prometheus-allow-scraping.ftl.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: allow-prometheus-scraping
+  namespace: ${namespace}
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              name: ${namePrefix}monitoring
```
schnatterer, line 12: Shouldn't that be ` kubernetes.io/metadata.name`?

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
```diff
@@ -145,7 +146,7 @@ class ApplicationConfigurator {
                     ],
                     springPetclinic: [
                             url: System.getenv('SPRING_PETCLINIC_REPO') ?: 'https://github.com/cloudogu/spring-petclinic.git',
-                            ref: '32c8653'
+                            ref: 'b0e0d18'
```
schnatterer, line 149: Please add a comment to the commit message that changes this that we update the petclinic version so it exposes prometheus metrics so we can easier test monitoring across namespaces.

schnatterer (changes requested):
We're almost there 👏

Please have a look at my minor findings and also add 
The reason to this commit message ` removing namespace additional labelling`: We learned about the built in ` kubernetes.io/metadata.name` label.

schnatterer (approved):

schnatterer:
Testing and merging on #228

PR #231, opened 2024-09-27
new maven image param and additional envs for docker run 

nihussmann:
It's now possible to set a new parameter: --maven-image altering the maven build logic in ces-build-lib. 
Additionally new docker run args can be passed down to jenkins.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
@@ -73,6 +73,9 @@ class ApplicationConfigurator {
                             version: '5.5.11'
                     ],
                     mavenCentralMirror: '',
+                    additionalEnv:  [
+                                    ADDITIONAL_DOCKER_RUN_ARGS: '-u0:0'
+                    ]
             ],
             scmm       : [
                     internal: true, // Set dynamically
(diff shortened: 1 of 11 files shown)


src/main/groovy/com/cloudogu/gitops/features/Jenkins.groovy
```diff
@@ -65,11 +65,18 @@ class Jenkins extends Feature {
                 INSTALL_ARGOCD            : config.features['argocd']['active'],
                 NAME_PREFIX               : config.application['namePrefix'],
                 INSECURE                  : config.application['insecure'],
-                URL_SEPARATOR_HYPHEN      : config.application['urlSeparatorHyphen']
+                URL_SEPARATOR_HYPHEN      : config.application['urlSeparatorHyphen'],
+                ADDITIONAL_ENV            : config.jenkins['additionalEnv']
```
schnatterer, line 69: I think we don't need that. These are passed to the init-jenkins.sh legacy shell script.

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
```diff
@@ -73,6 +73,9 @@ class ApplicationConfigurator {
                             version: '5.5.11'
                     ],
                     mavenCentralMirror: '',
+                    additionalEnv:  [
+                                    ADDITIONAL_DOCKER_RUN_ARGS: '-u0:0'
```
schnatterer, line 77: We really don't want to hard-code this! It's one example we might have to set via config, when installing GOP in an env where the agents are misconfigured or run a really old docker version.

nihussmann:
feature was split in separate PRs, see https://github.com/cloudogu/gitops-playground/pull/233 and https://github.com/cloudogu/gitops-playground/pull/234

PR #239, opened 2024-10-31
Fix netPols commit message

nihussmann:
fixing duplicates in commit message with active network policies 

Example:
image


schnatterer:
LGTM 
image

PR #240, opened 2024-11-04
Extend dev docs

schnatterer:
* Dev docs for two registries/airgapped/image pull secrets
* Feature.groovy to make developping new features easier

### DIFF
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -589,7 +589,7 @@ That's why we need to initialize our local cluster with some netpols for everyth
 After the cluster is initialized and before GOP is applied, do the following:
 
 ```bash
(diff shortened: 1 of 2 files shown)


schnatterer:
@ThomasMichael1811 as discussed in person, I took the liberty to elaborate on the hints you added on the language side.

I also tried to simplify the instructions a bit, by inlining the whole `helm upgrade` command for harbor.   
Hopefully this will make it easier in the future. I also considered putting all steps into a single script and even write it into a separate file like `docs/setupHarbor.sh` or similar. 
We could also keep that in mind for the next time we're on it.
What do you think?

PR #242, opened 2024-11-21
Prod image: Enhance GraalVM native image reliability

schnatterer:
Running more of the GOP initialization should avoid more `MethodNotFoundExceptions`.

Using heredoc in `Dockerfile` makes it easier to maintain.


schnatterer:
Skipping this PR, because heredoc (`EOF`) in `Dockerfile` is incomaptible with older docker versions that we would like to support.

All other changes of this PR are integrated in #236

PR #244, opened 2024-11-25
simpel Intellij code cleanup

nihussmann:
simpel Intellij code reformat and code cleanup. We should consider using Prettier in the future. For now this should work out. 

This branch is based on : feature/argocd-operator so merge this before. After this rebase again with main.


nihussmann:
we will create a new, better PR. .ftl files are kinda worse with Intellij

PR #247, opened 2024-12-11
Namespace isolation: Specify Prometheus NamespaceSelectors

nihussmann:
adding ServiceMonitorNamespaceSelectors to Prometheus.
removed duplicate getNamespace functions in Argo and Prometheus. Adding a Config param for active namespaces.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/Application.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/Application.groovy
@@ -1,6 +1,6 @@
 package com.cloudogu.gitops
 
-
+import com.cloudogu.gitops.config.Config
 import groovy.util.logging.Slf4j
 import jakarta.inject.Singleton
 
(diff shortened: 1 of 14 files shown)


src/main/groovy/com/cloudogu/gitops/Application.groovy
```diff
@@ -28,4 +33,31 @@ class Application {
...
+        String namePrefix = config.application.namePrefix;
+
+        if(config.registry.internal || config.scmm.internal || config.jenkins.internal){
+            namespaces.add(namePrefix + "default")
+        }
+        
+        if (config.features.argocd.active) {
+            namespaces.addAll(Arrays.asList(
+                    namePrefix + "argocd",
+                    namePrefix + "example-apps-staging",
```
ThomasMichael1811, line 48: if ArgoCd is active, example apps are mandatory?

ThomasMichael1811: Answer: Depends on old code. Just reused!

PR #248, opened 2024-12-16
Updated helm to v3.16.3

nihussmann:


schnatterer:
This ignores CVE-2024-45337 for four weeks, unit there hopefully is a helm version that fixes it

PR #249, opened 2024-12-16
BREAKING: Remove config options for Helm repo

ThomasMichael1811:
Currently it is not possible to change config for Helm repos using repoURL, chart and version.
So we remove this config option in order to avoid frustrated users.

Why?
In order to effectively use a different OCI or helm repo, we would have to add it to the AppProject.  Whilst this is not impossible, it is not implemented, yet. 
Same goes for authentication: If the different OCI repo would need credentials we would have to add those to the Argo CD config, which is also not implemented, yet.

So for now, we remove the option.
Once we have a requirement to implement it, we implemented with proper UX.

### DIFF
--- a/docs/configuration.schema.json
+++ b/docs/configuration.schema.json
@@ -188,17 +188,9 @@
                   "type" : [ "string", "null" ],
                   "description" : "Sets cainjector Image for Cert Manager"
                 },
-                "chart" : {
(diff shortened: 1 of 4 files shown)


ThomasMichael1811:
failing checks because of CVE in Scan Image step.

PR #251, opened 2024-12-19
Adds Helm values functionality to SCM Manager and other features

nihussmann:
- Introduced the `helmValues` function for SCM Manager to simplify configuration management.
- Set `initialDelaySeconds` directly in values giving more time for startup.
- Applied the same Helm values functionality to other features for consistency across the helm charts.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/features/Vault.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/features/Vault.groovy
@@ -20,11 +20,11 @@ import java.nio.file.Path
 class Vault extends Feature implements FeatureWithImage {
     static final String VAULT_START_SCRIPT_PATH = '/applications/cluster-resources/secrets/vault/dev-post-start.ftl.sh'
     static final String HELM_VALUES_PATH = 'applications/cluster-resources/secrets/vault/values.ftl.yaml'
-    
+
     String namespace = 'secrets'
     Config config
     K8sClient k8sClient
-    
+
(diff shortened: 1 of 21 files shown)


src/main/groovy/com/cloudogu/gitops/features/Vault.groovy
```diff
@@ -83,6 +80,7 @@ class Vault extends Feature implements FeatureWithImage {
             k8sClient.createNamespace('secrets')
             k8sClient.createConfigMapFromFile(vaultPostStartConfigMap, 'secrets', postStartScript.absolutePath)
 
+            //modifies the templatedMap Object directly, merges values for dev mode
```
ThomasMichael1811, line 83: instead of comment, you can use return value, like other features do.

src/main/groovy/com/cloudogu/gitops/features/Vault.groovy
```diff
@@ -121,11 +119,14 @@ class Vault extends Feature implements FeatureWithImage {
                                                     "/var/opt/scripts/${postStartScript.name} 2>&1 | tee /tmp/dev-post-start.log"
                                     ],
                             ]
-                    ], yaml)
+                    ], templatedMap)
         }
+        //merge custom helm values into chart
+        MapUtils.deepMerge(helmConfig.values,templatedMap)
```
ThomasMichael1811, line 125: use return value, like other features do.

PR #261, opened 2025-02-04
Petclinic: disable DB integrationstests

nihussmann:
Petclinic: Disable database integration tests because they start docker images (which won't work in air-gapped envs and take a lot of time in demos)


schnatterer:
This also fixes a failing unit test in petclinic, see cloudogu/spring-petclinic@994e6ff

PR #263, opened 2025-02-21
CLI: Simple params no longer require k8s

schnatterer:
`--help`, `--version` and `--output-config` should work even when there is no working kube-context.

e.g. 
```bash
docker run ghcr.io/cloudogu/gitops-playground:0.10.0 --version
```
Fails with:
```
couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
```

Because there is no working kube-context and initAndValidateConfig connects to k8s trying to find clusterBindAddress

This used to work before #241:
```
docker run ghcr.io/cloudogu/gitops-playground:0.8.0 --version
```

This PR fixes this issue.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
@@ -8,8 +8,8 @@ import ch.qos.logback.classic.spi.ILoggingEvent
 import ch.qos.logback.core.ConsoleAppender
 import com.cloudogu.gitops.Application
 import com.cloudogu.gitops.config.ApplicationConfigurator
-import com.cloudogu.gitops.config.schema.JsonSchemaValidator
 import com.cloudogu.gitops.config.Config
+import com.cloudogu.gitops.config.schema.JsonSchemaValidator
 import com.cloudogu.gitops.destroy.Destroyer
 import com.cloudogu.gitops.utils.CommandExecutor
 import com.cloudogu.gitops.utils.FileSystemUtils
@@ -17,12 +17,11 @@ import com.cloudogu.gitops.utils.K8sClient
 import groovy.util.logging.Slf4j
 import groovy.yaml.YamlSlurper
 import io.micronaut.context.ApplicationContext
-import jakarta.inject.Provider
 import org.slf4j.LoggerFactory
 import picocli.CommandLine
 
 import static com.cloudogu.gitops.config.ConfigConstants.APP_NAME
-import static com.cloudogu.gitops.utils.MapUtils.deepMerge
+import static com.cloudogu.gitops.utils.MapUtils.deepMerge 
 /**
  * Provides the entrypoint to the application as well as all config parameters.
  * When changing parameters, make sure to update the Config for the config file as well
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -44,24 +43,32 @@ class GitopsPlaygroundCli {
     ReturnCode run(String[] args) {
         setLogging(args)
 
-        def config = readConfigs(args)
-        def version = createVersionOutput()
+        log.debug("Reading initial CLI params")
```
ThomasMichael1811, line 46:
```suggestion
        log.debug("Reading initial CLI params {}", args)
```
if log info about CLI param, print them out too

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
```diff
@@ -44,24 +43,32 @@ class GitopsPlaygroundCli {
...
+        def cliParams = new Config()
```
ThomasMichael1811, line 47: Pathfinder, just think about it!
Currently only one command "about" GOP at same time is possible but multiple commands to configure GOP.

Suggestion like this for --version --help in one command:
```
boolean exit = false

        
        def version = createVersionOutput()
        if (cliParams.application.versionInfoRequested) {
            println version
//            return ReturnCode.SUCCESS
            exit = true;
        }

        def config = readConfigs(args)
        if (config.application.outputConfigFile) {
            println(config.toYaml(false))
            exit = true;
//            return ReturnCode.SUCCESS
        }

        if (cliParams.application.usageHelpRequested) {
            // if help is requested picocli help is used and printed by execute automatically
            new CommandLine(cliParams).execute(args)
//            return ReturnCode.SUCCESS
            exit = true;
        }

        if (exit){
            return ReturnCode.SUCCESS
        }
```
Now order is imporant, because print helpUsage changes cliParam to defaults

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
```diff
@@ -223,7 +221,7 @@ class ApplicationConfigurator {
         return newUrl
     }
 
-    private void validate(Config configToSet) {
+    void validateConfig(Config configToSet) {
```
ThomasMichael1811, line 224: Missing documentation for a public method

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
```diff
@@ -223,7 +221,7 @@ class ApplicationConfigurator {
...
         if (configToSet.scmm.url && !configToSet.jenkins.url ||
```
ThomasMichael1811, line 225: Pathfinder!
A method with a name like "isJenkinsSetWithSCM" could be helpful to understand quicker this boolean

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
schnatterer: Have you ever queried the version and help at the same time? 
At the moment `output-config` prints the YAML file only so you could pipe it into a YAML file. If it printed help or version before it would no longer be valid YAML.

IMO this makes everything more complex and would require more discussions than I intended. I merely wanted to restore the functionality that was lost during our changes.

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
schnatterer: This info already is being logged:
`[picocli INFO] Parsing 2 command line args [--help, --debug]`

ThomasMichael1811: perfect

src/main/groovy/com/cloudogu/gitops/cli/GitopsPlaygroundCli.groovy
ThomasMichael1811: ok

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
schnatterer: Implemented in a530d53

src/main/groovy/com/cloudogu/gitops/config/ApplicationConfigurator.groovy
schnatterer: Implemented in a530d53

PR #265, opened 2025-03-05
Fix integration tests for ingress-nginx

nihussmann:
Fixed two race condition issues:

- The integration test's start condition only checked for the presence of pods, while the test itself validated the Service. However, at that point, the Service was still in the process of being created.

- We waited for the node to return its name but did not wait for it to retrieve its IP. This occasionally caused flaky tests due to an empty IP value.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/utils/K8sClient.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/utils/K8sClient.groovy
@@ -13,6 +13,9 @@ import jakarta.inject.Singleton
 class K8sClient {
     private static final String[] APPLY_FROM_STDIN = ['kubectl', 'apply', '-f-']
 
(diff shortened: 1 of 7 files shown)


src/main/groovy/com/cloudogu/gitops/utils/K8sClient.groovy
```diff
@@ -28,13 +28,35 @@ class K8sClient {
     }
 
     String getInternalNodeIp() {
+
         String node = waitForNode()
 
-        // For k3d this is either the host's IP or the IP address of the k3d API server's container IP (when --bind-localhost=false)
```
schnatterer, line 33: Is this comment no longer valid?

schnatterer (changes requested):

schnatterer (approved):

PR #273, opened 2025-04-11
Namespace prefix for multi-tenancy support

nihussmann:


src/main/groovy/com/cloudogu/gitops/FeatureWithImage.groovy
```diff
@@ -15,12 +15,12 @@ trait FeatureWithImage {
...
 
-            log.trace("Creating image pull secret 'proxy-registry' in namespace ${namespace}")
+            log.trace("Creating image pull secret 'proxy-registry' in namespace ${this.namespace}")
             String url = config.registry.proxyUrl ?: config.registry.url
             String user = config.registry.proxyUsername ?: config.registry.readOnlyUsername ?: config.registry.username
             String password = config.registry.proxyPassword ?: config.registry.readOnlyPassword ?: config.registry.password
             
-            k8sClient.createNamespace(namespace)
+            k8sClient.createNamespace(this.namespace)
             k8sClient.createImagePullSecret('proxy-registry', namespace, url, user, password)
```
ThomasMichael1811, line 24: if "this" is neccessary for usage `this.namespace`, it should be used here, too.
```suggestion
            k8sClient.createImagePullSecret('proxy-registry', this.namespace, url, user, password)
```

PR #275, opened 2025-05-08
Adapt netpols to prefix

ThomasMichael1811:
Addition to #273

### DIFF
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -638,26 +638,21 @@ docker run --rm -t -u $(id -u) \
 The first increment of our `--netpols` feature is intended to be used on openshift and with an external Cloudogu Ecosystem.
 
 That's why we need to initialize our local cluster with some netpols for everything to work.
-* The `default` namespace needs to be accesible from outside the cluster (so GOP apply via `docker run` has access)
+* The `<prefix>-jenkins` and  `<prefix>-scm-manager` namespace needs to be accesible from outside the cluster (so GOP apply via `docker run` has access)
 * Emulate OpenShift default netPols: allow network communication inside namespaces and access by ingress controller 
 
 After the cluster is initialized and before GOP is applied, do the following:
 
 ```bash
+# Prefix handling:
+# if used, change prefix to your configured prefix and then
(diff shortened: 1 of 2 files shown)


docs/developers.md
```diff
@@ -638,26 +638,21 @@ docker run --rm -t -u $(id -u) \
 The first increment of our `--netpols` feature is intended to be used on openshift and with an external Cloudogu Ecosystem.
 
 That's why we need to initialize our local cluster with some netpols for everything to work.
-* The `default` namespace needs to be accesible from outside the cluster (so GOP apply via `docker run` has access)
+* The `<prefix>-jenkins` and  `<prefix>-scm-manager` namespace needs to be accesible from outside the cluster (so GOP apply via `docker run` has access)
```
schnatterer, line 641:
```suggestion
* The `<prefix>-jenkins` ,  `<prefix>-scm-manager` and `<prefix>-registry` namespace needs to be accesible from outside the cluster (so GOP apply via `docker run` has access)
```

docs/developers.md
```diff
@@ -690,6 +685,38 @@ spec:
...
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-all-ingress
+  namespace: ${prefix}registry
+spec:
+  podSelector: {}
+  ingress:
+  - {} 
+EOF
+
```
schnatterer, line 719:
```suggestion
# Some NS need to be accessible from docker image
for ns in ${prefix}jenkins ${prefix}registry ${prefix}scm-manager; do
  k apply --namespace "$ns" -f- <<EOF
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: allow-all-ingress
spec:
  podSelector: {}
  ingress:
  - {}
EOF
done
```

schnatterer: I think a loop is much easier to read, shorter and also consistent to the loop above

schnatterer (commented):
Great that your thorough testing lead to finding this bug!
I've proposed some polishing.

PR #285, opened 2025-06-12
Update gitops-playground-production.drawio.svg

NicoSien:
changed background color to white


flxebrt:
Added a light background which appears light in the dark mode as well

PR #307, opened 2025-10-01
add recent changes for upstream contribution

mdroll:
Add openshift flavor specific configuration to upstream

### DIFF
--- a/argocd/argocd/operator/argocd.ftl.yaml
+++ b/argocd/argocd/operator/argocd.ftl.yaml
@@ -8,7 +8,7 @@ spec:
     enabled: true
     resources:
       limits:
-        cpu: '2'
+        cpu: 500m
         memory: 1Gi
       requests:
         cpu: 250m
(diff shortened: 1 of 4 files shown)


argocd/argocd/operator/argocd.ftl.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     enabled: true
     resources:
       limits:
-        cpu: '2'
+        cpu: 500m
```
nihussmann, line 11:
```suggestion
        cpu: '2'
```

argocd/argocd/operator/argocd.ftl.yaml
```diff
@@ -39,7 +39,7 @@ spec:
   controller:
     resources:
       limits:
-        cpu: "2000m"
+        cpu: "500m"
```
nihussmann, line 42:
```suggestion
        cpu: "2000m"
```

PR #375, opened 2026-03-29
Feature/oci via contentloader

ThomasMichael1811:
Implement OCI registries for helm charts


ThomasMichael1811:
not neccessary anymore

PR #467, opened 2026-04-24
Password for admin is now randomized and printed after successful start.

ThomasMichael1811:
- Admin password is now generated
- introduce new GOP start option: --gop-namespace 

if --gop-namespace is used, GOP stores initial admin password and whole configuration in this namespace as secret.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/config/Config.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/config/Config.groovy
@@ -442,9 +442,11 @@ class Config {
 		Boolean clusterAdmin = false
 
 		@Option(names = ["-p", "--profile"], description = APPLICATION_PROFIL)
+		@JsonPropertyDescription(APPLICATION_PROFIL)
 		String profile
 
 		@Option(names = ["--gop-namespace"], description = APPLICATION_GOP_NAMESPACE)
+		@JsonPropertyDescription(APPLICATION_GOP_NAMESPACE)
 		String gopNamespace = ''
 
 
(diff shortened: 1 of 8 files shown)


src/main/groovy/com/cloudogu/gitops/config/Config.groovy
```diff
@@ -56,7 +58,7 @@ class Config {
 	// When updating please also adapt in Dockerfile, vars.tf and init-cluster.sh
 	public static final String K8S_VERSION = "1.35.4"
 	public static final String DEFAULT_ADMIN_USER = 'admin'
-	public static final String DEFAULT_ADMIN_PW = 'admin'
+	public static final String DEFAULT_ADMIN_PW = generatePassword()
```
DerDaehne, line 61: gop will create a random password on every run, even if the user specifically set a initial password via config file/profile. 
In that case, the applications will be configured with the user-provided password, but gop will still print the random password on the welcomeScreen as well as store it in the gop-namespace (if that functionallity is enabled), which may lead to confusion on the user's side.

src/main/resources/application-dev.yaml
```diff
@@ -0,0 +1,74 @@
...
+      overwriteMode: UPGRADE
+
+  namespaces:
+    - ${config.application.namePrefix}example-apps-production
+    - ${config.application.namePrefix}example-apps-staging
+  variables:
+    petclinic:
+      baseDomain: "petclinic.localhost"
+    nginx:
+      baseDomain: "nginx.localhost"
```
DerDaehne: nginx examples have been removed from gop, so these config values are unnecessary.

DerDaehne:
LGTM

PR #481, opened 2026-05-11
moves K8sClient to Fabric8 Lib

nihussmann:
Refactored the K8sClient to Fabric8. Removed dependencies


ThomasMichael1811 (commented):
all OK

PR #488, opened 2026-05-15
argocd operator install

nihussmann:
install the argocd operator with a config flag


DerDaehne (approved):

DerDaehne:
The changes proposed in this PR are not really necessary for our current use-cases. If we were to merge this, we would have to re-introduce a command executor for kubectl, which we not really want to do. 
i'm closing this pr for now, but we'll keep the branch around.

PR #492, opened 2026-05-28
Feature/integrationtests logs

FelixWende99:
- Improve integration test failure diagnostics
- New Awaitility helpers in src/test/groovy/com/cloudogu/gitops/integration/TestK8sHelper.groovy:144 for pods, pod prefix lists, matchers, and namespaces.
- ensureIngressIsOnline and the other FullProfileTestIT checks now use waits instead of single-shot assertions: src/test/groovy/com/cloudogu/gitops/integration/profiles/FullProfileTestIT.groovy:85.
- Similar flaky spots in PrefixProfileTestIT, PetclinicProfileTestIT, MandantProfileTestIT, and ArgoCDProfileTestIT have been converted.
- Hardcoded pod counts were removed or replaced, for example for external-secrets and cert-manager. Cert-Manager now checks the expected components instead of exactly 3 pods: src/test/groovy/com/cloudogu/gitops/integration/features/CertManagerTestIT.groovy:50.
- Monitoring now also explicitly waits for the operator, not just indirectly for Grafana: src/test/groovy/com/cloudogu/gitops/integration/features/MonitoringTestIT.groovy:51.

also:
The helper now detects fatal pod/container states such as `Failed`, `Succeeded`, `CrashLoopBackOff`, `ImagePullBackOff`, `ErrImagePull`, and non-zero terminated container exits. Instead of waiting for the full Awaitility timeout, affected tests now fail immediately with a more useful error message.

For prefix- and matcher-based checks, the behavior still allows rollout scenarios where one matching pod is unhealthy but another matching pod is already `Running`.


ThomasMichael1811 (approved):
ok

PR #493, opened 2026-05-28
This PR fixes several issues in the Fabric8-based Kubernetes client and ArgoCD operator deployment flow.

avetgit:
- Added proper support for custom resources by resolving CRD metadata and using `ResourceDefinitionContext` instead of calling `genericKubernetesResources` with only a resource string.
 - Fixed recursive RBAC YAML application by removing the unsupported `--recursive` CLI-style argument and applying YAML files from directories directly.


src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
```diff
@@ -179,7 +179,7 @@ class ArgoCD extends Tool {
 		log.debug("Apply RBAC permissions for ArgoCD in all managed namespaces imperatively")
 		// Apply rbac yamls from operator/rbac folder
 		String argocdRbacPath = clusterResourcesRepo.operatorRbacDir()
-		k8sClient.applyYaml("${argocdRbacPath} --recursive")
```
nihussmann, line 182: why is this removed?

avetgit: --recursive is unsupported

nihussmann (approved):

PR #501, opened 2026-06-11
Add an option to configure the listen loadbalancers http/https listen hosts

DerDaehne:
Some users might want to setup a demo environment to be accessible from the local network. 
The current script will bind all cluster ingresses to localhost. While this is a very good default setting, allowing it to be confgurable provides more flexibility.

This PR will introduce a new parameter "--bind-ingress-host" to do just that.


ThomasMichael1811 (approved):
LGTM

PR #505, opened 2026-06-16
Prevent SCM-Manager Argo CD Tracking Collisions in Multi-Tenant Setups

avetgit:
## Pull request overview

Adjusts SCM-Manager/Argo CD naming and bootstrap behavior to avoid Argo CD tracking collisions in prefixed/multi-tenant setups, and fixes Kubernetes ConfigMap retrieval to use the client’s active namespace.

**Changes:**
- Update `K8sClient.getConfigMap()` to read ConfigMaps from `client.namespace` (fallback to default).
- Prefix SCM-Manager Helm/ArgoCD release/service naming using `config.application.namePrefix`, and update related URL resolution + tests.
- Change Argo CD Application generation to (a) prefix `metadata.name` to avoid tracking collisions and (b) special-case SCM-Manager as a bootstrap app (embed Helm values; no self-referencing values git source); also set Argo CD resource tracking method to `annotation`.

### DIFF
--- a/argocd/cluster-resources/apps/argocd/argocd/values.ftl.yaml
+++ b/argocd/cluster-resources/apps/argocd/argocd/values.ftl.yaml
@@ -70,6 +70,7 @@ argo-cd:
     cm:
       timeout.reconciliation: 15s
       repository.check.interval: 30s
+      application.resourceTrackingMethod: annotation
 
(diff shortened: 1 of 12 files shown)


mdroll (approved):
LGTM

PR #512, opened 2026-06-24
This PR updates K8sClient to resolve custom resources via the Kubernetes Discovery API 

ThomasMichael1811:
## Pull request overview

This PR updates `K8sClient` to resolve custom resources via the Kubernetes Discovery API (`/apis`, `/apis/<group>/<version>`) instead of listing `CustomResourceDefinitions`, reducing the need for cluster-wide CRD list permissions.

**Changes:**
- Replace CRD-list-based custom resource resolution with Discovery API-based resolution and build `ResourceDefinitionContext` from discovered group/version/resource metadata.
- Introduce `KubernetesApiResourceNotFoundException` for cases where an API resource cannot be resolved via discovery.
- Add tests to verify discovery-based resolution for ArgoCD CRs and the exception behavior when resolution fails.


FelixWende99 (approved):
Looks good

PR #516, opened 2026-06-26
Fix SCM-Manager URL resolution for tenant and central providers

avetgit:
Resolve SCM-Manager service names and namespaces using a provider-specific service prefix instead of the tenant application's global name prefix.

Previously, the SCM-Manager URL resolver always used `config.application.namePrefix` when building internal service URLs. In dedicated multi-tenant setups this caused the central SCM-Manager provider to be resolved with the tenant prefix, producing URLs such as `t3-scmm.t3-scm-manager.svc.cluster.local` instead of the central `scmm.scm-manager.svc.cluster.local`.

This broke tenant bootstrap generation because central Argo CD tried to load the tenant cluster-resources repository from the tenant SCM-Manager instead of the central SCM-Manager.

Tenant SCM providers now receive the tenant service prefix, while central SCM providers are resolved independently. This keeps tenant SCM URLs prefixed and central SCM URLs stable.

### DIFF
--- a/src/main/groovy/com/cloudogu/gitops/application/orchestration/GitHandler.groovy
+++ b/src/main/groovy/com/cloudogu/gitops/application/orchestration/GitHandler.groovy
@@ -101,7 +101,7 @@ class GitHandler {
 					config.multiTenant.scmManager,
 					k8sClient,
 					networkingUtils,
-					'')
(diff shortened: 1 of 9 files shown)


ThomasMichael1811 (approved):
lgtm

PR #522, opened 2026-07-02
Introduce DeploymentOrchestrator for tool execution

FelixWende99:
## Summary

Introduces a central DeploymentOrchestrator and aligns repository provisioning with the deployment context flow. The DeploymentContext is now created once in Application.start(), passed explicitly through the orchestration flow, and no longer injected into tool or deployment singletons. Repository provisioning prepares one RepositoryWorkspace, and the orchestrator passes that workspace to each active tool.

## Changes

- Added DeploymentOrchestrator as the central orchestration component.
- Moved explicit tool deployment order into the orchestrator.
- Updated Application to build DeploymentContext once and hand it to repository provisioning and tool orchestration.
- Added Tool.execute(context, workspace) as the explicit orchestrator entry point.
- Moved the previous install flow directly into Tool.execute(context, workspace).
- Removed the separate Tool.install() / enabled-install path from the common tool lifecycle.
- Kept isEnabled(context) as a pure activation check without deployment preparation or config mutation.
- Moved tool state preparation, such as namespace setup, into prepare() hooks in the tool execution flow.
- Updated RepositoryProvisioning to receive DeploymentContext explicitly in workflow methods.
- Updated RepositoryProvisioning to choose single-instance vs dedicated-instance workspace from context.isMultiTenant().
- Updated ArgoCD, ScmManager, Monitoring, Jenkins, Registry, Ingress, CertManager, ExternalSecretsOperator, Vault, and ContentLoader to use the orchestrator-provided context/workspace.
- Removed DeploymentContext constructor injection from destruction handlers and aligned the destroy flow with the runtime context lifecycle.
- Updated tests for orchestrator ordering, context/workspace handover, tool execution, and deployment strategy wiring.

## Verification
  - `git diff --check` passed.
  - `./mvnw -Dtest=DeployerTest,ArgoCdApplicationStrategyTest,IngressTest,JenkinsTest,RegistryTest,MonitoringTest,ExternalSecretsOperatorTest,VaultTest,CertManagerTest,ScmManagerSetupTest,ApplicationTest test` passed.
  - `./mvnw -Dtest=ApplicationTest,DestroyerDependencyInjectionTest,JenkinsApiClientTest,JobManagerTest test` passed.


src/main/groovy/com/cloudogu/gitops/application/Application.groovy
```diff
@@ -47,14 +49,10 @@ class Application {
...
+		RepositoryWorkspace workspace = repositoryProvisioning.provideWorkspace()
 
-		tools.forEach(tool -> {
-			tool.validate()
-		})
-
-		tools.forEach(tool -> {
-			tool.install()
-		})
+		deploymentOrchestrator.execute(context,
```
avetgit: I would rename deploymentOrchestrator.execute(context, workspace) to
deploymentOrchestrator.deployTools(context, workspace)

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+	DeploymentOrchestrator(List<Tool> tools) {
+		this.tools = tools
+	}
+
+	void execute(DeploymentContext context,
+		RepositoryWorkspace workspace) {
+		log.debug('Starting tool orchestration.')
+
+		tools.each { Tool tool ->
```
avetgit, line 52: First, the orchestrator should not simply run through all tools unconditionally. It should only execute the tools that are actually active according to the config or the tool activation logic. Otherwise, disabled tools may still be validated or executed.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
```diff
@@ -0,0 +1,66 @@
...
+			log.debug("Validating tool ${tool.class.simpleName}")
+			tool.validate()
```
avetgit: I would not call validate() separately from the orchestrator. The ticket prepares a unified entry point with execute(context, workspace), which can later encapsulate the lifecycle validate -> preDeploy -> deploy -> postDeploy. If the orchestrator already calls validate() itself, it becomes aware of an internal lifecycle phase and part of the lifecycle control moves outside the tool.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		tools.each { Tool tool ->
```
avetgit: Something like this:
void execute(DeploymentContext context, RepositoryWorkspace workspace) {
    log.debug('Starting tool orchestration.')

    tools
        .findAll { Tool tool -> tool.isEnabled(context) }
        .each { Tool tool ->
            log.debug("Executing tool ${tool.class.simpleName}")
            tool.execute(context, workspace)
        }

    log.debug('Tool orchestration finished.')
}

FelixWende99: Done

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
FelixWende99: Done

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
FelixWende99: Done

src/main/groovy/com/cloudogu/gitops/application/Application.groovy
FelixWende99: Done

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
avetgit, line 1: I think `GitHandler` currently takes over responsibility that should stay in `RepositoryProvisioning` or be read directly from the `DeploymentContext`.

`multiTenantDeployment` is not really Git-/SCM-specific information. It is a general runtime condition of the GOP run and is already available through `context.isMultiTenant()`. If `GitHandler` stores this value and `RepositoryProvisioning` later asks `gitHandler.isMultiTenantDeployment()`, the `GitHandler` effectively becomes a proxy for the deployment context.

src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
avetgit, line 1: From my point of view, `RepositoryProvisioning` should decide between the single-instance and dedicated-instance workspace directly based on the `DeploymentContext`:

```groovy
if (context.isMultiTenant()) {
    workspace = createDedicatedInstanceWorkspace()
} else {
    workspace = createSingleInstanceWorkspace()
}

Comment (avetgit) [inline src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy]: I would prefer passing the context explicitly to the workflow methods, for example:

```groovy
class RepositoryProvisioning {

    static final String CLUSTER_RESOURCES_REPO_TARGET = 'argocd/cluster-resources'

    private final GitRepoFactory gitRepoFactory
    private final GitHandler gitHandler

    private RepositoryWorkspace workspace
    private boolean remoteRepositoriesEnsured = false
    private boolean repositoriesCloned = false

    RepositoryProvisioning(GitRepoFactory gitRepoFactory,
                           GitHandler gitHandler) {
        this.gitRepoFactory = gitRepoFactory
        this.gitHandler = gitHandler
    }

    void prepare(DeploymentContext context) {
        ...
    }

    RepositoryWorkspace provideWorkspace(DeploymentContext context) {
        ...
    }
}

File: src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
Comment (FelixWende99) [reply src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy:1]: Agreed. I removed the deployment-mode state from GitHandler; it now only prepares/selects SCM providers. Deployment mode is read from DeploymentContext at the call site.

File: src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
Comment (FelixWende99) [reply src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy:1]: done

Comment (FelixWende99) [reply src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy]: Agreed and changed. prepare(context) and provideWorkspace(context) now receive DeploymentContext explicitly, so RepositoryProvisioning no longer stores context as singleton state.

File: src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
Comment (avetgit) [inline src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy]: `DeploymentContext` should not be injected into tool constructors anymore. It is created once per GOP run in `Application.start()`.
- Tools should use the `RepositoryWorkspace` passed through `tool.execute(context, workspace)`.

File: src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy [hunk 5]
```diff
@@ -82,7 +82,7 @@ class ArgoCD extends Tool {
 
 	@Override
 	void enable() {
-		this.repositoryWorkspace = repositoryProvisioning.provideWorkspace()
+		this.repositoryWorkspace = repositoryProvisioning.provideWorkspace(context)
```
avetgit: - `ArgoCD` should not call `repositoryProvisioning.provideWorkspace(context)` itself, because that bypasses the orchestrator-provided workspace and couples the tool back to repository provisioning.

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
avetgit: - `ArgoCD` also has its own `repositoryWorkspace` field, although `Tool` already provides one. This should be avoided to prevent field shadowing.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+	DeploymentOrchestrator(List<Tool> tools) {
+		this.tools = tools
+	}
+
+	void deployTools(DeploymentContext context,
+		RepositoryWorkspace workspace) {
+		log.debug('Starting tool orchestration.')
+
+		tools.findAll { Tool tool -> tool.isEnabled(context)
```
avetgit: - We should also align the enablement flow: either the orchestrator filters with `isEnabled(context)` or `Tool.install()` checks enablement, but not both with different method signatures.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
```diff
@@ -0,0 +1,62 @@
...
+		}.each { Tool tool ->
+			log.debug("Executing tool ${tool.class.simpleName}")
+			tool.execute(context,
```
avetgit: I would prefer the orchestrator to pass the prepared workspace into each active tool, and the tools should work with that workspace only. This keeps repository provisioning centralized and prevents tools from resolving or recreating their own workspace.

src/main/groovy/com/cloudogu/gitops/application/context/DeploymentContextFactory.groovy
avetgit, line 1: I would avoid adding a `DeploymentContextFactory` that exposes `DeploymentContext` as a singleton bean.

`DeploymentContext` represents the state of a specific GOP run. If we register it as a Micronaut singleton again, services can inject it through their constructors and we are back to a globally managed context object. That also makes the creation time harder to control and can lead to the context being built before the Config has been fully prepared.

I would prefer keeping only `ContextBuilder` as a singleton service and building the context explicitly in `Application.start()`:

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
FelixWende99: Resolved. Tool constructors no longer receive DeploymentContext; it is created once in Application.start() and passed through tool.execute(context, workspace).

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
FelixWende99: Resolved. ArgoCD now uses the RepositoryWorkspace passed by the orchestrator and no longer requests a workspace from RepositoryProvisioning.

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
FelixWende99: Resolved. ArgoCD uses the inherited Tool.repositoryWorkspace field now, avoiding field shadowing.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
FelixWende99: Resolved. The orchestrator calls tool.execute(context, workspace) for the configured flow, while the common install behavior stays in Tool.installEnabledTool() to avoid duplicated install logic.

src/main/groovy/com/cloudogu/gitops/application/orchestration/DeploymentOrchestrator.groovy
FelixWende99: Resolved. DeploymentOrchestrator passes the prepared RepositoryWorkspace into every tool via execute(context, workspace), and downstream deployment also receives that workspace through Deployer.

src/main/groovy/com/cloudogu/gitops/infrastructure/git/GitRepoFactory.groovy
```diff
@@ -8,15 +8,13 @@ import jakarta.inject.Singleton
...
 	protected final FileSystemUtils fileSystemUtils
 
-	GitRepoFactory(DeploymentContext context, FileSystemUtils fileSystemUtils) {
+	GitRepoFactory(FileSystemUtils fileSystemUtils) {
 		this.fileSystemUtils = fileSystemUtils
-		this.context = context
 	}
 
-	GitRepo create(String repoTarget, GitProvider gitProvider) {
+	GitRepo create(DeploymentContext context, String repoTarget, GitProvider gitProvider) {
```
avetgit: GitRepo doesn't nee context. GitRepo should use Config config instead

src/main/groovy/com/cloudogu/gitops/application/content/ContentLoader.groovy
```diff
@@ -329,7 +327,7 @@ class ContentLoader extends Tool {
 		if (repoConfig.templating) {
 			def engine = getTemplatingEngine()
 
-			GitRepo repo = this.repoProvider.create(repoConfig.target, this.gitHandler.tenant)
+			GitRepo repo = this.repoProvider.create(context, repoConfig.target, this.gitHandler.tenant)
```
avetgit: GitRepo doesn't need context

src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
```diff
@@ -167,16 +141,17 @@ class RepositoryProvisioning {
 		return CLUSTER_RESOURCES_REPO_TARGET
 	}
 
-	private RepositoryWorkspace createSingleInstanceWorkspace() {
+	private RepositoryWorkspace createSingleInstanceWorkspace(DeploymentContext context) {
 		log.debug('Creating single-instance repository workspace.')
 
-		GitRepo clusterResourcesRepository = gitRepoFactory.create(clusterResourcesRepoTarget(),
+		GitRepo clusterResourcesRepository = gitRepoFactory.create(context,
```
avetgit: GItRepo doesn't need context

src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
```diff
@@ -185,10 +160,12 @@ class RepositoryProvisioning {
 		 * Therefore both repositories are represented explicitly, even though they use the same
 		 * logical repository target.
 		 */
-		GitRepo clusterResourcesRepository = gitRepoFactory.create(clusterResourcesRepoTarget(),
+		GitRepo clusterResourcesRepository = gitRepoFactory.create(context,
```
avetgit: same here, context is not needed here

src/main/groovy/com/cloudogu/gitops/application/context/DeploymentContextFactory.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/infrastructure/deployment/ArgoCdApplicationStrategy.groovy
```diff
@@ -198,6 +225,12 @@ class ArgoCdApplicationStrategy implements DeploymentStrategy {
 			"Add ${repoName}/${chartOrPath} to ArgoCD")
 	}
 
+	void useRuntime(DeploymentContext context,
```
avetgit: This method is only used in Test. It should be removed

src/main/groovy/com/cloudogu/gitops/utils/AirGappedUtils.groovy
```diff
@@ -38,15 +39,16 @@ class AirGappedUtils {
 	 *
 	 * @return the repo namespace and name
 	 */
-	String mirrorHelmRepoToGit(HelmConfig helmConfig) {
+	String mirrorHelmRepoToGit(DeploymentContext context,
```
avetgit: We don't nee context here

src/main/groovy/com/cloudogu/gitops/tools/common/Tool.groovy
```diff
@@ -129,7 +125,7 @@ abstract class Tool {
 		if (context.isAirgapped()) {
 			log.debug("Using a local, mirrored git repo as deployment source for feature ${featureName}")
 
-			String repoNamespaceAndName = this.airGappedUtils.mirrorHelmRepoToGit(helmConfig)
+			String repoNamespaceAndName = this.airGappedUtils.mirrorHelmRepoToGit(context, helmConfig)
```
avetgit: We don't need context here

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
```diff
@@ -26,37 +25,43 @@ import org.springframework.security.crypto.bcrypt.BCrypt
...
 	private ArgoCDRepoSetup repoSetup
 	private ArgoCDRepoLayout clusterResourcesRepo
 
-	ArgoCD(DeploymentContext context,
-		K8sClient k8sClient,
+	ArgoCD(K8sClient k8sClient,
 		HelmClient helmClient,
 		FileSystemUtils fileSystemUtils,
 		GitHandler gitHandler,
 		RepositoryProvisioning repositoryProvisioning) {
```
avetgit: RepositoryProvisioning repositoryProvisioning can be removed from ArgoCD COnstructor

src/main/groovy/com/cloudogu/gitops/infrastructure/git/GitRepoFactory.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/application/content/ContentLoader.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/application/repository/RepositoryProvisioning.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/utils/AirGappedUtils.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/tools/common/Tool.groovy
FelixWende99: removed

src/main/groovy/com/cloudogu/gitops/tools/core/argocd/ArgoCD.groovy
FelixWende99: Done

src/main/groovy/com/cloudogu/gitops/infrastructure/deployment/ArgoCdApplicationStrategy.groovy
FelixWende99: Done

avetgit (approved):
LGTM :)

PR #536, opened 2026-07-08
Move tool-specific GitOps resource preparation into owning tools

avetgit:
## Summary

This PR continues the repository preparation refactoring by moving tool-specific GitOps resource handling out of the central ArgoCD setup and into the owning tools.

Instead of having `ArgoCDRepoSetup` prepare resources for multiple tools, each tool is now responsible for preparing its own `apps/<tool>` content in the `cluster-resources` repository. ArgoCD keeps responsibility for ArgoCD-owned resources only.

## Motivation

Previously, repository preparation responsibilities were centralized in `ArgoCDRepoSetup`. This made the class responsible for both ArgoCD-specific setup and GitOps resources owned by other tools.

This PR improves the separation of responsibilities:

* Tool-specific GitOps resources are prepared by the owning tool.
* ArgoCD-specific resources remain in the ArgoCD setup.
* Repository changes are coordinated through the shared `RepositoryWorkspace`.
* Deployment still runs through the existing deployment flow.

This makes the code easier to understand, reduces central special handling, and provides a cleaner foundation for future tool-specific changes.

## Changes

* Move tool-specific GitOps resource preparation into the respective tool classes.
* Limit `ArgoCDRepoSetup` to ArgoCD-owned repository preparation.
* Use `RepositoryWorkspace` as the shared access point for cluster resources and tenant bootstrap repositories.
* Render copied templates before they are consumed by ArgoCD.
* Commit and push generated GitOps resource changes through `RepositoryWorkspace`.
* Pass `DeploymentContext` and `RepositoryWorkspace` through the deployment flow.
* Adjust deployment strategies, the deployer, and related tests to use the updated method signatures.
* Update tests to verify the new ownership and repository preparation behavior.

## Notes

The intended deployment behavior stays the same. The main change is the ownership of the preparation logic: each tool now prepares its own GitOps resources instead of relying on the central ArgoCD setup.


src/main/groovy/com/cloudogu/gitops/tools/Ingress.groovy
```diff
@@ -53,7 +60,26 @@ class Ingress extends Tool implements ToolWithImage {
 
 	@Override
 	void enable() {
+		prepareIngressApp(repositoryWorkspace.clusterResourcesRepository)
+
 		def helmConfig = config.features.ingress.helm
-		deployHelmChart('traefik', 'traefik', namespace, helmConfig, HELM_VALUES_PATH, context)
+
+		deployHelmChart(TOOL_NAME,
```
FelixWende99, line 67: The workspace path and the ArgoCD values source diverge here. The tool prepares `apps/ingress`, but `deployHelmChart()` receives `traefik` as the feature/repo name. `ArgoCdApplicationStrategy` derives `toolPath = apps/${repoName}`, so it writes and references `apps/traefik` instead.

Potential fix: pass the repository workspace subfolder explicitly, or use a consistent logical tool name for the workspace path while keeping `traefik` only as the Helm release/application name where needed.

src/main/groovy/com/cloudogu/gitops/tools/CertManager.groovy
```diff
@@ -53,6 +59,24 @@ class CertManager extends Tool implements ToolWithImage {
 
 	@Override
 	void enable() {
-		deployHelmChart('cert-manager', 'cert-manager', namespace, config.features.certManager.helm, HELM_VALUES_PATH, context)
+		prepareCertManagerApp(repositoryWorkspace.clusterResourcesRepository)
```
FelixWende99, line 62: `prepareCertManagerApp()` copies `apps/cert-manager`, including `misc/cluster-issuers.ftl.yaml`, but the copied templates are not rendered afterwards. Since the ArgoCD Application references the tool folder recursively, this can leave FreeMarker syntax in the GitOps repository and make ArgoCD try to apply invalid YAML.

Potential fix: render the copied cert-manager templates after copying, similar to Vault/Monitoring, and add a test that `apps/cert-manager/misc/cluster-issuers.yaml` exists while the `.ftl.yaml` file does not.

see `src/main/groovy/com/cloudogu/gitops/tools/Vault.groovy:114`
  ```
private void replaceVaultTemplates(GitRepo clusterResourcesRepo) {
      clusterResourcesRepo.replaceTemplates([config: config])
  }
```

src/main/groovy/com/cloudogu/gitops/tools/core/scmmanager/ScmManager.groovy
FelixWende99: The bootstrap push happens before `createArgocdApplication()`. This means the initial cluster-resources push after SCM-Manager deployment does not contain the SCM-Manager ArgoCD Application; it is pushed in a second commit afterwards. That seems to contradict the intended bootstrap flow, where the initial push should contain the prepared repository state plus SCM-Manager generated artifacts.

Potential fix: create/write the SCM-Manager ArgoCD Application before the bootstrap commit/push, or move the bootstrap commit so it happens after all SCM-Manager GitOps artifacts have been written.

src/main/groovy/com/cloudogu/gitops/tools/Ingress.groovy
avetgit: Thanks, good point.

apps/ingress is the source path for the Ingress-owned GitOps resources. traefik is the deployment target name, because the Ingress tool deploys the Traefik Helm chart/release.

src/main/groovy/com/cloudogu/gitops/tools/CertManager.groovy
avetgit: Done :)

src/main/groovy/com/cloudogu/gitops/tools/core/scmmanager/ScmManager.groovy
avetgit: I changed that and separate SCM-Manager bootstrap repository preparation from the final push step. 
This ensures generated GitOps artifacts, such as the SCM-Manager ArgoCD Application, are written before the initial bootstrap state is committed and pushed.

src/main/groovy/com/cloudogu/gitops/tools/CertManager.groovy
FelixWende99: Could we rename `replaceVaultTemplates` to `replaceCertManagerTemplates`?

FelixWende99 (approved):
lgtm!

PR #540, opened 2026-07-14
Migrate to Gradle as build tool

DerDaehne:
This PR migrates the project from Maven to Gradle as its build tool. This lays the foundation for a broader codebase transition, moving from Groovy to Kotlin.


DerDaehne:
we will stick with maven for now. closing

PR #549, opened 2026-08-12
Decouple deployable tools from the central Config

avetgit:
## Summary

This PR decouples deployable tools from the central GOP `Config` by introducing dedicated, tool-specific configuration DTOs and mappers.

Previously, tools could access the complete configuration through `DeploymentContext`, for example via `context.getConfig()`. While this was a pragmatic transitional solution during the introduction of `DeploymentContext`, `DeploymentOrchestrator`, repository provisioning, and the common tool lifecycle, it also meant that every tool could depend on arbitrary parts of the global configuration.

With this refactoring, each tool now receives only the configuration values it actually needs.

## What changed

* Introduced dedicated `ToolConfig` DTOs for deployable tools.
* Added corresponding `ToolConfigMapper` implementations.
* Introduced `AbstractMappedTool<T>` to map tool configuration before the tool lifecycle starts.
* Replaced direct `context.getConfig()` access in deployable tools with `toolConfig()`.
* Moved interpretation of central configuration values into the respective mapper.
* Reduced dependencies on central `Config` types and constants inside tool implementations.
* Added focused configuration representations for commonly used Helm and image pull secret settings.
* Projected template configuration into dedicated plain data structures instead of passing central `Config` schema objects through the DTO boundary.
* Added defensive immutable copies for nested configuration maps and collections while preserving valid `null` values in Helm values.
* Kept the configuration lifecycle explicitly separated from the deployment lifecycle through `ConfigLifecycleHook`.
* Preserved existing tool behavior while making configuration dependencies explicit and testable.

## Architecture

The configuration flow for deployable tools now follows this structure:

```text
DeploymentContext + Config
          |
          v
 ToolConfigMapper<T>
          |
          v
 Tool-specific DTO
          |
          v
 AbstractMappedTool<T>
          |
          v
 Tool lifecycle
```

The tool lifecycle therefore no longer needs knowledge about the structure of the complete GOP configuration.

Configuration mapping and lifecycle execution remain separate responsibilities.

The configuration lifecycle is also intentionally independent from the deployment lifecycle:

```text
Deployment lifecycle
    AbstractTool
        |
        v
    AbstractMappedTool<T>

Config lifecycle
    ConfigLifecycleHook
```

Only components that actually participate in configuration initialization implement `ConfigLifecycleHook`.

## Benefits

This refactoring provides a clearer architectural boundary between the central GOP configuration and individual tools.

It:

* reduces coupling between tools and the global configuration model,
* makes configuration dependencies of each tool explicit,
* prevents accidental access to unrelated configuration areas,
* improves isolated testability of tool logic,
* localizes future changes to the central `Config` primarily in mapper implementations,
* separates configuration interpretation from tool behavior,
* provides a cleaner basis for future refactorings of the configuration model.

## ContentLoader

`ContentLoader` is intentionally not migrated to the tool-specific DTO mechanism in this PR.

Although it currently also depends on the central configuration, it is not a deployable tool and has broader responsibilities than the components covered by this refactoring.

Its responsibilities and its long-term configuration boundary should therefore be evaluated separately instead of forcing it into the tool abstraction introduced here.

The general handling and architectural decomposition of `ContentLoader` will be addressed in a dedicated follow-up.

## Out of scope

This PR does not:

* change the structure of the central `Config`,
* introduce new configuration options,
* intentionally change functional tool behavior,
* redesign `ContentLoader`,
* introduce a general domain-wide configuration model.

## Testing

The existing tool tests were adapted to the DTO-based configuration boundary.

Additional tests cover, among other things:

* mapping of relevant configuration values into tool DTOs,
* correct mapping before lifecycle execution,
* fail-fast behavior for invalid mapper setup,
* defensive copying of nested configuration values,
* preservation of `null` values in Helm configuration,
* explicit mapping of tool-specific configuration semantics such as Vault development mode,
* preservation of existing behavior for configuration edge cases.

## Result

Deployable tools now depend on a focused configuration contract instead of the complete GOP configuration.

This makes the configuration boundary explicit and establishes a more maintainable foundation for future changes to both the central configuration model and the tool lifecycle architecture.


ThomasMichael1811 (approved):
more or less ok.

PR #553, opened 2026-08-17
Remove Config from DeploymentContext

avetgit:
This PR removes the global Config from DeploymentContext and turns the context into a snapshot of interpreted deployment-wide decisions only.

Config dependencies are now passed explicitly to the components that actually need them. Git infrastructure components were decoupled from DeploymentContext where possible, while ContentLoader was intentionally kept out of scope except for the minimal wiring required to preserve existing behavior.

This improves dependency transparency, reduces hidden coupling through DeploymentContext, and creates a cleaner boundary between configuration and deployment context.


src/main/java/com/cloudogu/gitops/infrastructure/deployment/ArgoCdApplicationTargetResolver.java
```diff
@@ -3,16 +3,19 @@
...
 
 import java.util.regex.Pattern;
 
 @Singleton
+@RequiredArgsConstructor
 public class ArgoCdApplicationTargetResolver {
 
 	private static final Pattern TRAILING_DASH = Pattern.compile("-$");
 
+	private final Config config;
```
ThomasMichael1811, line 16: how this class gets config Object?

src/main/java/com/cloudogu/gitops/infrastructure/git/providers/scmmanager/ScmManagerUrlResolver.java
```diff
@@ -17,35 +15,39 @@ public class ScmManagerUrlResolver {
...
+	private final String repositoryNamePrefix;
+	private final boolean runningInsideK8s;
 	private final String servicePrefix;
 
 	private URI cachedClusterBind;
 
-	public ScmManagerUrlResolver(DeploymentContext context, ScmManagerConfig scmm, K8sClient k8s, NetworkingUtils net) {
-		this(context, scmm, k8s, net, "");
+	public ScmManagerUrlResolver(
+		ScmManagerConfig scmm,
```
ThomasMichael1811: should have "config" in name

src/main/java/com/cloudogu/gitops/infrastructure/kubernetes/rbac/RbacDefinition.java
```diff
@@ -21,7 +21,7 @@ public class RbacDefinition {
 	private List<ServiceAccountRef> serviceAccounts = new ArrayList<>();
 	private String subfolder = "rbac";
 	private GitRepo repo;
-	private Config config;
+	private Map<String, Object> config;
```
ThomasMichael1811, line 24: not sure, if own class make more sense. Lets discuss

src/main/java/com/cloudogu/gitops/infrastructure/kubernetes/rbac/Role.java
```diff
@@ -12,7 +10,7 @@ public record Role(
 
 	Variant variant,
 
-	Config config
+	Map<String, Object> config
```
ThomasMichael1811, line 13: same: ownClass instead Map ?

src/main/java/com/cloudogu/gitops/tools/common/AbstractMappedTool.java
```diff
@@ -0,0 +1,60 @@
...
+	protected AbstractMappedTool(ToolConfigMapper<T> toolConfigMapper) {
+		this.toolConfigMapper = Objects.requireNonNull(toolConfigMapper, "Tool config mapper must not be null");
+	}
+
+	@Override
+	public final boolean isEnabled(DeploymentContext context) {
+		return isEnabled(mapConfig(context));
+	}
+
+	protected abstract boolean isEnabled(T config);
```
ThomasMichael1811: having an order would be nice. 
start with variables,
privates, protecteds,
then abstract methods, 
then non-abstract-methods

would be easier to read.

src/main/java/com/cloudogu/gitops/infrastructure/kubernetes/rbac/RbacDefinition.java
avetgit: TemplateConfig is needed for freemarker templating, so we don't use the complete application config

src/main/java/com/cloudogu/gitops/tools/common/AbstractMappedTool.java
avetgit: done

src/main/java/com/cloudogu/gitops/infrastructure/kubernetes/rbac/Role.java
avetgit: We use this map for freemarker templating

src/main/java/com/cloudogu/gitops/infrastructure/kubernetes/rbac/RbacDefinition.java
avetgit: We use this map for freemarker templating

src/main/java/com/cloudogu/gitops/infrastructure/git/providers/scmmanager/ScmManagerUrlResolver.java
avetgit: done

src/main/java/com/cloudogu/gitops/infrastructure/deployment/ArgoCdApplicationTargetResolver.java
avetgit: Config is injected by Micronaut through constructor injection. @RequiredArgsConstructor generates a constructor for the final config field.

ThomasMichael1811 (approved):

PR #590, opened 2026-09-17
Respect selected integration test profile on manual main builds

avetgit:
Manual builds on main currently ignore the selected chooseProfile value and execute all integration test profiles.

This change keeps the existing behavior for scheduled and automatic main builds, while respecting the selected profile for manually triggered builds.

Examples:

operator-full → runs only operator-full
full → runs only full
all-profiles → runs all profiles

The existing develop behavior remains unchanged.


mdroll (approved):
lgtm

